PR #1, opened 2017-11-01
allow any and all descriptors to be used in calibration matching rule…

anulance:
…s (avoids the need to update the static descriptor list in cal_service whenever a descriptor useful for, or required by, calibration matching is added)

### DIFF
--- a/recipe_system/cal_service/calrequestlib.py
+++ b/recipe_system/cal_service/calrequestlib.py
@@ -15,21 +15,12 @@ from gempy.utils import logutils
 from .caches  import set_caches
 from recipe_system.cal_service import cal_search_factory, handle_returns_factory
 from .file_getter import get_file_iterator, GetterError
+from astrodata import descriptor_list
 # ------------------------------------------------------------------------------
 log = logutils.get_logger(__name__)
 # ------------------------------------------------------------------------------
 # Currently delivers transport_request.calibration_search fn.
 calibration_search = cal_search_factory()
-# ------------------------------------------------------------------------------
-descriptor_list = ['amp_read_area', 'camera', 'central_wavelength', 'coadds',
-                   'data_label', 'data_section', 'detector_roi_setting',
-                   'detector_x_bin', 'detector_y_bin', 'disperser',
-                   'exposure_time', 'filter_name', 'focal_plane_mask',
-                   'gain_setting', 'gcal_lamp', 'instrument', 'lyot_stop',
-                   'nod_count', 'nod_pixels', 'object', 'observation_class',
-                   'observation_id',
-                   'observation_type', 'program_id', 'read_speed_setting',
-                   'ut_datetime', 'read_mode', 'well_depth_setting']
 # ------------------------------------------------------------------------------
 def get_request(url, filename):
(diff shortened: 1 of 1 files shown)


KathleenLabrie:
Looks good to me.  @chris-simpson mentioned that the "except AttributeError" probably needs to be changed to a more encompassing error, and then have the value set to None.   But then I see that in the code right now it's a "try-except AttributeError-else".   

Can @chris-simpson and @kiloRomeoAlpha check the error handling please?  Now that we are returning all the descriptors, some are bound to be un-applicable and crash.  For the calibration request, those should just return a None for the query.

chris-simpson:
I'm not sure the "except AttributeError" is required as the descriptor must exist for it to be in descriptor_list (which is compiled from a list of suitably-decorated methods). I think we should create a "DescriptorError" for when a descriptor is called inappropriately (e.g., dispersion_axis for an image) and this is what should be caught by the "try-except" when the descriptor is called (which is currently a global try-except).

KathleenLabrie:
@anulance FYI, the plan is for Chris to create that DescriptorError first, before we merge.

anulance:
@KathleenLabrie Sounds good.  Yes, I (briefly) thought maybe the exception handling could go away, as all descriptors returned should by definition be present.  But I didn't think to protect against descriptors that shouldn't be present!  It would be cool if descriptors could be defined based on tag set so only appropriate descriptors would be present/returned... but I'm sure that requires significant refactoring.

kiloRomeoAlpha:
Has anyone sent a cal request with the full descriptor list to the Calibration manager? My larger concern is that pushing previously unseen descriptor values to the calibration manager does not cause problems. They probably won't unless one (or more) descriptors result in some data type FitsStorage does not expect. Python being python, this probably won't be a problem unless FitsStorage calls a method on a data type that does not have that attr. Did anyone mention this to Paul?

kiloRomeoAlpha:
Looks like I was being a worry wort. calibrations requests FitsStorage with the full descriptor list appears to cause _no_ problems.

kiloRomeoAlpha:
BTW, @anulance, there is an import bug in this calrequestlib module. We recently moved the caches.py module contents to geminidr. I didn't catch this, as there was still a .pyc file in my dir masking the problem. But can you fix this line in this same module, calrequestlib for the merge? Thanks!

Line 15 cal_requestlib.py:
---------------------------
from .caches  import set_caches
TO:
from geminidr  import set_caches

chris-simpson:
It transpires that none of the descriptors raise exceptions any more -- they all return None if called inappropriately (e.g., dispersion_axis on an image). Catching the AttributeError should be unnecessary but it does no harm, so this is good to go.

anulance:
@kiloRomeoAlpha, @chris-simpson  Sorry I just missed the boat: Chris merged the request only a few minutes ago... I didn't get a chance to update line 15.  Did you want me to issue another pull request, or do you want to handle it on your end?

chris-simpson:
My bad. I should have done that before merging. I've pushed a fix direct to
master so we're good.

anulance:
Cool

PR #2, opened 2017-11-15
Rectified bug that would cause crash if stackFrames called without apply_dq argument

marc-white:
The provided default parameters for GHOST slit frame stacking failed
to provide a default value for the apply_dq option to stackFrames (this
has now been fixed). This resulted in the RecipeSystem crashing at the
line modified in this commit, as the key 'apply_dq' did not exist in
the params dictionary. The new get call will return None if the key is
not defined (although as good practice, it probably should be anyway).

### DIFF
--- a/gempy/gemini/eti/gemcombineparam.py
+++ b/gempy/gemini/eti/gemcombineparam.py
@@ -86,7 +86,7 @@ class Masktype(GemcombineParam):
     def __init__(self, inputs=None, params=None):
         log.debug("Masktype __init__")
         GemcombineParam.__init__(self, inputs, params)
-        if params["apply_dq"]:
+        if params.get("apply_dq"):
             self.masktype = "goodvalue"
         else:
             self.masktype = "none"


KathleenLabrie:
Looks fine to me.  Let's have someone else check it out too.

chris-simpson:
"apply_dq" is the parameter previously known as "mask", so it wasn't missing, it has been renamed. This was because stackSkyFrames() now has a "mask_objects" parameter and confusion would abound. So the "mask" parameter in the stackFrames() parameter set is no longer needed and you should delete that.

marc-white:
Two questions/points on that:
1. Would you be wanting the crash behavior for all missing parameters, or just those where there 
isn't a sensible and/or obvious default? For example, if the user hasn't explicitly said, "Yes, I want the DQ plane applied," you can go ahead and just not do it without bombing the reduction - would that be a safe assumption?
2. In the case where you do want the system to raise an error for a missing parameter, would we want/is there already a more specific error that could be raised, as opposed to a fairly generic KeyError?

chris-simpson:
1. The user doesn't have to explictly say anything about any parameters. The defaults are defined in the parameters_xxx.py modules, and they *must* be defined -- if they're not defined there, this is an error and something should barf. Now, apply_dq=True in parameters_stack.py (so, in fact, the opposite of what you suggest the default behaviour should be), so let's suppose we have the code adopt that as the default if it can't find the parameter. Now suppose you want to have the GHOST default be False yet you mistakenly type "apply_sq: False" in your parameter file; you'll think you're running with the parameter set to False by default, but it's actually doing what the code has defaulted to (i.e., True). My general view is that, if there's a mistake, code shouldn't try to guess what you want.

2. Just to reiterate, you were getting an error because we changed the name of a parameter on a primitive where you have overridden the default parameter set, so your par set dict had a mis-named key. Had I realized this, I would have told you and this wouldn't have arisen -- this was the result of a major refactor of the near-IR reduction primitives and I hope we're not going to make further changes with such wide-ranging impact. The error you saw can *only* arise because of a malformed par set dict and so (assuming we distribute working software) absolutely *cannot* happen in general operations but only for users who set about creating their own modules containing new parset dicts. I think such people will cope with a KeyError, which does specify the name of the missing key so is rather informative.

PR #4, opened 2018-12-05
Updated astrodata tests to either run with local data or skip tests.

b1quint:
astrodata's tests are hardcoded using some reference data. If other data is used for any reason, they fail. To start changing that and considering that we want to have a local and an online test scenarios, I changed its tests to add the "--ad_test_data_path" to the astrodata's tests. If this option is not set, the tests are skipped with the proper skip message.


b1quint:
Tests were limited to run only on `astrodata`, `recipe_system` and `gemini_instruments`.

PR #8, opened 2019-01-04
RTD can't read several modules. Fix attempt

b1quint:


b1quint:
Won't fix this for now. I know that when RTD uses Python 2.7, several docs are not documented. But when it uses Python 3.5, everything works file.

PR #16, opened 2019-02-04
Flamingos 2 Tutorial

b1quint:
Added the Flamingos2 Tutorial inside `geminidr/docs/tutorials`. This tutorial is using some packages that were added to the `requirements.txt` file (like ìpython`, `matplotlib` and `cython`).


jehturner:
I'm not sure how far along this tutorial is and have only had the briefest look at it for now, but I don't think we want to be telling users to install DRAGONS from a zip file... We've gone to some effort to produce conda packages and the expectation is that users will just do something like `conda install iraf-all pyraf-all stsci gemini` or maybe `conda install stsci dragons`. Also, regarding terminology, the thing you install into is usually referred to as a "conda environment" rather than a "virtual environment", I think to avoid confusion with `virtualenv` (which is, or at least was, incompatible with Anaconda). Thanks!

jehturner:
Regarding the documentation build, I have a sphinx conda package that generates the documentation when building our conda Python packages (though it might want things like the theme tweaking). You cannot install sphinx that way yourself, due to some obscure path length problem, but it gets the HTML and/or PDF into `share/` in the built conda packages. Maybe it would be easiest if we try to set up our conda package build in Jenkins in the near future, so you don't have to worry about this stuff (depending on your workflow)? We need to re-establish our continuous integration for package builds anyway and we should be running tests against the *actual* package(s) that we eventually distribute to users (maybe not always but at some point).

b1quint:
You are right about using `conda` to install DRAGONS instead of using a zip file. I remembered that Astrodata User's Manual and the Recipe System User's Manual had those but I completely blanked out that information from my head. I will remove the installation step from this doc (and the other tutorials) and take into consideration that `conda` has to be used as the preferred installation method.

jehturner:
Thanks. It doesn't *have* to be used, but it is certainly the preferred installation method.

PR #40, opened 2019-07-26
Doc/gmos tutorial

b1quint:
- Updated GMOS tutorials with images processed with the most recent version of `reduce`.
- Added sections with guidelines on how to generate Processed Fringe Frames.
- Added appendix with a table with Detector/Filters that require Processed Fringe Frames for reduction. It might be safe to check this with the version I have in my local branch when I am back.


b1quint:
If the Goodman Tutorial is good we can delete the branch.

PR #42, opened 2019-11-22
Serialize wavelength/distortion calibrations as gWCS

jehturner:
~Do not merge (though I believe this initial AstroData commit is harmless & working).~ This PR is for discussing a first implementation of wavelength calibration propagation as gWCS.

In the longer run, some more design work will be needed in order to accommodate co-ordinate transformations in general, which is outside the scope of this PR (except insofar as it's needed to get the immediate application working).

I'm currently pushing a subset of my code that actually does the serialization in AstroData, since that stands by itself. It should support arbitrary gWCS objects, as long as the constituent models have "tags" defined in `astropy.io.misc.asdf`, which means using `astropy` `master` (or 4.0rc) to get Chebyshev models. This also works to some extent with AstroPy 3, as long as you don't try to use files generated with AstroPy 4 (which actually load the WCS with a warning but then it doesn't work properly; I should probably add a check to drop the WCS in that case).

This is the original concept using YAML in ASCII tables, as conceived with STScI about 5 years ago (before they did something else), but we might want to use a more-common binary table instead, which should also be able to store the text. One of my reasons for sticking with ASCII in the first instance is that Chris wanted the solution to be human-readable, which, in principle, ASCII is to a greater extent than a FITS-specific binary format (if you can extract the extension). In practice, however, there are more tools for working with binary tables, eg. in IRAF, so a `BinTableHDU` might be more accessible in that sense, which might be what Chris meant. Anyway, I wanted to demonstrate this option.

@chris-simpson @rcardenes @saimn


jehturner:
Yeah, the meaning of life, the Universe and everything. I always knew it (at least it feels that way after slowly planning this since 2011).

astrodata/fits.py
```diff
@@ -326,6 +330,95 @@ def normalize_indices(slc, nitems):
...
+    The ASCII table is actually a mini ASDF file. The constituent AstroPy
+    models must have associated ASDF "tags" that specify how to serialize them.
+    """
+
+    import jsonschema
+
+    # Create a small ASDF file in memory containing the WCS object
+    # representation because there's no public API for generating only the
+    # relevant YAML subsection and an ASDF file handles the "tags" properly.
+    try:
```
jehturner: Some of this code should probably be spun off into non-FITS-specific routines to support serializing gWCS in other formats, but for now it works.

astrodata/fits.py
```diff
@@ -326,6 +330,95 @@ def normalize_indices(slc, nitems):
...
+                        .format(extver or ''))
+
+    # ASDF can only dump a YAML to a binary file object, so do that and read
+    # the text back from it for storage in a FITS extension:
+    with BytesIO() as fd:
+        with af:
+            # Generate the YAML, dumping any binary arrays as text:
+            af.write_to(fd, all_array_storage='inline')
+        fd.seek(0)
+        wcstext = fd.read().decode()
```
jehturner: Some of this code should probably be spun off into non-FITS-specific routines to support serializing gWCS in other formats, but for now it works.

astrodata/nddata.py
```diff
@@ -372,3 +374,15 @@ def transpose(self):
...
                              mask=None if self.mask is None else self.mask.T, copy=False)
         return new
+
+    @property
+    def wcs(self):
+        return super().wcs
+
+    @wcs.setter
+    def wcs(self, value):
+        allowed = (WCS,)
```
jehturner: Argh, I forgot to include `None` here.

astrodata/nddata.py
```diff
@@ -372,3 +374,15 @@ def transpose(self):
...
+        if not isinstance(value, allowed):
+            raise ValueError("wcs value must be instance of: {}"\
```
jehturner: This should probably be a `TypeError`.

jehturner: This should probably be a TypeError.

jehturner:
Yes to FITS WCS (though in the long run it might be preferable to use `gWCS` for images too?) and (at some point) FITS approximation. However, there was no drawback to adding support for just `gWCS` in the meantime, since I don't think we've implemented FITS WCS on `NDData` and it therefore does everything it did before. But yes, I should probably add `wcs.WCS` to the list of allowed types, along with `NoneType`... This might constrain our API to APE14, however, raising the question of how to deal with intermediate gWCS co-ordinate frames. Maybe the assumptions about the WCS object will just be context dependent in primitives. The bookeeping for co-ordinate frames already needs more thought in the near future anyway.

Any thoughts on the ASCII vs binary table? One slight concern I have is what will happen if ASDF *does* write a few binary bytes for some reason (which could cause it to fail). I asked Nadia to confirm whether `all_array_storage='inline'` guarantees that this won't happen and she hasn't replied to that. I think it should do, but writing bytes to a binary table might be more robust in that respect. On the other hand, I'd have to check whether existing tools can be coaxed to present the data to the user in some sensible format (and not as, say, a long column of ASCII code digits). But not having it ever fail is obviously a priority at the level of AstroData.

chris-simpson:
What does the ASCII serialization look like? If it's not readable (and your message suggests it's not), it may as well be binary. Even if it is readable, it'll be almost impossible to edit (compared to an IRAF database file, which can be edited) but it has some debugging value. I think I probably need to do some playing around with this branch to understand it better. But the important thing is that we can do this.

jehturner:
In IRAF, `tables` fails to read this serialization of the distortion model, crashing with a memory error, but it appears that it actually *can* work with ASCII tables after all and the extension type was a red herring. The error seems more like a problem with the length of the field. Although `io.fits` handles the long string without any problems and I *think* it is FITS compliant, this is not really surprising given that IRAF is full of data structures with hard-coded limits (not to mention bugs). The other issue is that `tread` & `tedit` do not format the long string containing line breaks well, making it unnavigable in the editor. I think both of these problems can largely be solved by splitting the text into one line per table row, which is probably more natural anyway, though the YAML structure then needs preserving carefully (doesn't look difficult).

Even after doing that, the `tables` editor strips the indentation, which is syntactically meaningful, and its navigation is a bit mangled if ones tries to edit the entries (cursor in the wrong place type of bug). The input table does have the correct structure; it just doesn't get presented correctly for this purpose (`tprint` & `tdump` do a better job). It is possible to edit a YAML entry as intended with a bit of artful fiddling, but then the indentation of that row is lost permanently when saving it, corrupting the structure. So this serialization perhaps has to be considered read-only in IRAF -- which seems reasonable given that no-one is actually going to be *using* gWCS in IRAF -- it's just one means of inspecting the files.

In Python -- with or without the lines split -- there's no problem extracting the text using `io.fits`, of course, though it's pretty fiddly for users. The current version with line breaks in a single row is slightly easier to extract the YAML from, since you don't have to rejoin the lines afterwards, but that seems a minor consideration compared with keeping _some_ level of read compatibility with IRAF (and any other library that might conceivably choke on long strings). In AstroData, which is a more realistic user interface, we could make it easy to see the YAML if we really wanted to, but that's probably duplicating the job of the gWCS API, including it's `__repr__`, which I think provides a comparable summary of the models.

In, say, a text editor, the ASCII version is slightly more readable than the binary one because there are no special characters within the table, but splitting the lines into (fixed-width) rows eliminates the line breaks in both cases, causing staggered column alignment and difficult navigation, as for the FITS headers (and those aren't quite as bad because they're 80 characters). So it's semi-readable, but I don't think that's an option to lose much sleep over. Actually editing FITS is not going to work if it changes the file size.

Perhaps one of your concerns is how the archive will present the information in its Web interface? There's no reason why it couldn't tack the WCS YAML onto the existing header dump that one of the links goes to, but presumably AstroData would then need to provide the archive with access to that text somehow, rather than just reconstructing the WCS object. I'm thinking the best interface in principle may be to (re)generate that YAML on the fly when requested, so it's always in sync with the object, and also use that to produce the serialization -- but for the archive you really want to show the contents of the input file with as little processing as possible, so that suggests some unfortunate duplication. Anyway, it should be no problem to display an understandable representation in the archive, however we decide to do it.

In either type of table extension, storing non-ASCII characters in a character column seems to violate the standard, so I don't think a binary table helps us there, if we want it to be readable. We could probably encode any special characters as ASCII (under the reasonable assumption that there won't be many of them, nor any fixed pointers within the file), if we want to be paranoid. So I think ASCII vs a binary table doesn't matter much (and is almost the same code), but it seems marginally preferable to stick to ASCII.

I don't see that any of this is much less user-accessible than your existing `WAVECAL`/`FITCOORDS` binary tables (except for not being able to edit coefficients in IRAF) -- right?  But we gain the ability to serialize almost any chain of transformations we want, as well as a standardized interface etc. If we really wanted the text to be trivially editable like an IRAF database file, that **could** also be arranged, but I think we'd need to pull the YAML out of FITS into some separate file, which is what we're trying to get away from. We could also have a tool to allow editing/reading/writing it, but that's better done through the programmatic interface (remembering that no such option exists in IRAF). So I think the human-readability limitations are as much an issue with storing metadata in FITS as anything specific to gWCS.

Sorry this got so long...

jehturner:
I think this is now as human readable as it's going to get, @chris-simpson (though I'll push an actual example in the next batch of changes)...

I've also modified it to fall back to saving a binary table (with a warning) if it fails to convert the ASDF for some arbitrary gWCS object to ASCII. I believe that should **not** happen given the way I'm using ASDF, but it seems best to ensure that our software doesn't simply fail at a low level -- and the resulting binary table would be useful for debugging how non-ASCII data got in there -- and it's still easy to reproduce an understandable gWCS object from **if** you have `AstroData` installed. If we **really** don't want such things in the archive, we could always add an ingestion check to reject the files and alert us to the problem. But again, I don't think this is going to happen. I did also look at options for encoding any stray binary as inline text ourselves, but found that not as easy as it sounds, I think principally because it's hard to do unambiguously without making breakable assumptions. In general, I've tried fairly hard to avoid making more fragile assumptions here than necessary.

Regarding the new dependencies in this PR, `geminidr` (and hence DRAGONS) is going to require `astropy >4.0` & `asdf >2.4` (which are Python 3 only) in order to serialize wavelength distortions etc. However, I'm wary of breaking `AstroData` itself on Python 2; right now it's not distributed stand-alone, but it could (and probably should) be and I'm not sure whether we've already made decisions that would break `AstroData` on 2 in `master`? As things currently stand in this PR, `AstroData` now requires `asdf` (only when there's a gwcs object) & `gwcs` (whether or not there is one) but they can be *any* versions of those packages -- it may fail to serialize certain `gwcs` objects if the versions are too old, but that's a generally-unavoidable hazard. I don't think our code itself should break with old versions, since the `asdf`/`gwcs` API usage is very simple and has probably been the same all along, so it will just do its best. So if we wanted to, we could retain Python 2 compatibility for `AstroData` by allowing installation of old `asdf`, `astropy` & `gwcs` versions and everything will be at least as good as before (even reading files with a wavelength WCS produced by Python 3 should just give a warning). But the full DRAGONS conda package will require the new versions.

jehturner:
The test for this will fail until https://github.com/astropy/astropy/issues/9809 is fixed.

astrodata/fits.py
```diff
@@ -326,6 +329,130 @@ def normalize_indices(slc, nitems):
...
+        # bytes that ASDF can parse. If AstroData has produced another format,
+        # it will be a binary dump due to the unexpected presence of non-ASCII
+        # data, in which case we just extract unmodified bytes from the table.
+        if colarr.dtype.kind in ('U', 'S'):
+            sep = os.linesep
+            # Not sure whether io.fits ever produces 'S' on Py 3, but just in
+            # case: join lines as str & avoid a TypeError with unicode linesep;
+            # could also use astype('U') but it assumes an encoding implicitly.
+            if colarr.dtype.kind == 'S' and not isinstance(sep, bytes):
+                colarr = np.vectorize(lambda b: b.decode('ascii'))(colarr)
```
saimn: `np.chararray.decode` provides a simpler way for this.

astrodata/fits.py
```diff
@@ -326,6 +329,130 @@ def normalize_indices(slc, nitems):
...
+            return
+
+        # If this table column contains text strings as expected, join the rows
+        # as separate lines of a string buffer and encode the resulting YAML as
+        # bytes that ASDF can parse. If AstroData has produced another format,
+        # it will be a binary dump due to the unexpected presence of non-ASCII
+        # data, in which case we just extract unmodified bytes from the table.
+        if colarr.dtype.kind in ('U', 'S'):
+            sep = os.linesep
+            # Not sure whether io.fits ever produces 'S' on Py 3, but just in
```
saimn: There is a parameter in `fits.open` to control this: `character_as_bytes`, default is `False`.

astrodata/fits.py
```diff
@@ -1039,6 +1166,9 @@ def to_hdulist(self):
             if ext.mask is not None:
                 hlst.append(new_imagehdu(ext.mask, header, 'DQ'))
 
+            if isinstance(ext.wcs, WCS):
+                hlst.append(wcs_to_asdftablehdu(ext.wcs, extver=ver))
```
saimn: What happens if `wcs` is an astropy `WCS` object, not a `gwcs` one ?
For consistency and more general use, would it be possible to fallback on loading the astropy `WCS` from the header if the `(g)wcs` extension is missing?

jehturner: Yes, @csimpson and I have discussed that and just haven't got to it yet. In the meantime, this adds functionality that doesn't already exist... My main purpose here was to demonstrate that we can serialize a `gwcs` attribute and get the use case for spectroscopy working.

jehturner: The idea is additionally to keep a FITS approximation of the gWCS solution so that eg. display tools can show useful co-ordinates, but that obviously requires more work. I've also discussed this a bit with Nadia recently and they are planning to use a look-up table for that purpose, which (although highly redundant) makes some sense, since it doesn't actually need to be an approximation or difficult to generate.

astrodata/fits.py
jehturner: Done. There seems to be some problem with using `np.vectorize` on a `chararray` anyway, which puzzles me a bit since I tested this stuff, but I might have tested it a bit out of context without knowing the `io.fits` option to get a byte array.

astrodata/fits.py
jehturner: I've modified the comment slightly.

jehturner:
It looks like the only test failures here were from the long slit sky subtraction, which is expected due to an issue in `master`.

jehturner:
This will be merged via the `gwcs_for_spect` branch.

PR #45, opened 2020-01-27
Refac/populate provenance data

olyoberdorf:
Moved branch/PR within github for convenience


chris-simpson:
In recipe_system.utils.decorators.py

_get_provenance_inputs():

Nothing is added if ad.filename=None. Generally, all ADs should have a filename but if one doesn't, then the failure to even add a placeholder will casue the list returned by this function to be shorter and get out of step with the adinputs list.

The md5sum is calculated if ad.path is not None, but ad.path might point to a file that no longer exists (again, it shouldn't happen but might), in which case it returns None, rather than the "" you apparently want, so you maybe just want to do something like:
md5 = md5sum(ad.path) or ""

_clone_provenance():

(We just discussed) This checks whether a provenance record is already in the provenance and doesn't call add_provenance() if it is, but add_provenance() also does that check. What you're doing could be more efficient if most records already exist in the destination AD's provenance but I doubt there's much difference and it's just extra code.


In astrodata.core.py

lines like:
        if hasattr(self._dataprov, 'GEM_PROVENANCE_HISTORY'):
            provenance_history = self._dataprov.GEM_PROVENANCE_HISTORY

you can drop the "._dataprov". __getattr__(ad, 'THING_IN_CAPITALS') will go and look in the DataProvider for you; I guess there's a minor efficiency gain in going there directly but this is dependent on the current way AD works and that might change.

            self.append(my_astropy_table, name='GEM_PROVENANCE')

should be self.GEM_PROVENANCE = my_astropy_table (stylistically, I don't like the name "my_astropy_table"  and you can just do "self.GEM_PROVENANCE = Table(as_created_in_line_above)". This one actually matters because we're likely to limit what "append()" can do when AD is revamped.

Also, why the "GEM_" in "GEM_PROVENANCE" and "GEM_PROVENANCE_HISTORY"? I assume it's short for "Gemini" but it's being added regardless isn't it?

        primitive = np.array([ph.primitive for ph in history]) etc.

is there any need to make an array, rather than just the list? "primitive = [ph.primitive for ph in history]" should be create a list (rather than a generator) that Table() will work with.

Your choice of timestamp string "%Y-%m-%d %H:%M:%S.%f" should probably be abstracted to a CONSTANT defined at the top of the module to avoid repetition.


In various primitives (this example is specifically primitives_ccd)

lines like:
            ad.add_provenance(Provenance(datetime.now(), bias.filename, md5sum(bias.path), "biasCorrect"))

here you again have the issue that md5sum() might return None rather than the null string you apparently want. Also, the static string "biasCorrect" should be replaced with self.myself() so it's robust to a change in primitive name.

olyoberdorf:
Made changes per @chris-simpson 's suggestions

PR #49, opened 2020-02-18
1D and 2D spectra registration

saimn:


geminidr/gemini/lookups/timestamp_keywords.py
```diff
@@ -11,23 +11,23 @@
     "addObjectMaskToDQ": "ADOBJMSK",
     "addReferenceCatalog": "ADDRECAT",
     "addVAR": "ADDVAR",
+    "adjustWCSToReference": "CORRWCS",
+    "adjustSlitOffsetToReference": "CORRSLOFF",
```
saimn: @chris-simpson - Is that ok for the timestamp keyword (`CORRSLOFF`) ?

geminidr/core/parameters_spect.py
```diff
@@ -102,6 +102,38 @@ def validate(self):
         if self.w1 is not None and self.w2 is not None and self.w2 <= self.w1:
             raise ValueError("Ending wavelength must be greater than starting wavelength")
 
+
+class adjustSlitOffsetToReferenceConfig(config.Config):
+    suffix = config.Field("Filename suffix", str, "_slitOffsetCorrected",
```
saimn: @chris-simpson - I used `_slitOffsetCorrected` for the suffix, unless you have a better suggestion ?

geminidr/gemini/lookups/timestamp_keywords.py
saimn: Oups, it seems I'm having trouble counting to eight :). Let's go for `CORRSLOF` then ?

chris-simpson:
I haven't run the code; you have a test so I assume things work as expected (e.g., the sign of the offset is correct). We generally work under the philosophy that the most successful tests of new code are generally to just push it and let it break as people start to use it in anger.

A couple of connected points: in resampleToCommonFrame() you only check for the adjustSlitOffsetToReference() timestamp keyword for the second and later ADs, and let the first AD go through assuming it has a slit offset of zero. I think it would be better to explicitly add SLITOFF=0 to the reference AD in adjust() and check it in resample() (and use its SLITOFF value). It looks like your coding requires that the first AD in adjust() is also the first AD in resample(), which is true in "normal" operations where a recipe is run, but need not be in other circumstances. If SLITOFF=0 is added to the first AD in adjust(), then the resampling will work if any subset of the ADs are sent to resample(), and in any order. This also makes the looping easier in that primitive as you can loop over all adinputs (instead of adinputs[1:]), avoiding most of the code for the first pass of the loop, and add the timestamping to the loop.

Also, please put the adjust() primitive at the top of the module, in alphabetical order, so it can be found (I scrolled through a few times trying to find it!).

saimn:
Ok, thanks for the feedback. I added SLITOFF=0 for the reference ad, which indeed simplifies a bit things later and make it more flexible :+1: 
Should those primitives go in the recipes ?

PR #50, opened 2020-02-19
refac/upload science

olyoberdorf:
I'll want to retest this, but making the PR early so @chris-simpson can take a look when he has time


chris-simpson:
storeProcessedScience() is missing a "self." before "log". The subsequent logic looks strange. If the mode isn't recognized, it says it's not saving anything, but continues anyway: should the second "if" be an "elif"? Then, storeCalibration() looks like it will save the file into the "calibrations/" directory and I don't think we want that (unless you've been told otherwise). Looking at how much code in storeCalibration() isn't needed for science (creating the "calibrations/" directory if it doesn't exist, saving the file there, etc.) I think all we need is the "try: upload_calibration" block, so it's more efficient to just dump that directly in storeProcessedScience().

PR #51, opened 2020-02-21
Add warning when user param is ignored because the recipe enforce it

saimn:
Following something we discussed a while ago: if a parameter is enforced by the recipe than it is silently ignored if the user set it to a different value. Adding a warning in this case.


jehturner:
Looks good at a glance. Chris is really the logger man though.

PR #53, opened 2020-02-21
Implement gWCS serialization in the primitives

jehturner:
This PR on the currently-unmerged PR #42 is for changes to primitives that might break the reduction, whereas #42 itself is just for core changes to `AstroData` to support `gWCS` serialization and should be safe to merge in the near future. Got it?


jehturner:
@chris-simpson

jehturner:
(Merged recent changes from `master` via `gwcs` so both branches are up to date now.)

PR #62, opened 2020-04-30
Add option to save a rejection map

saimn:
Will add a few comments inline to explain some points.
ping @chris-simpson @jehturner

### DIFF
--- a/astrodata/fits.py
+++ b/astrodata/fits.py
@@ -1624,6 +1624,8 @@ def windowedOp(func, sequence, kernel, shape=None, dtype=None,
     finally:
         astropy.log.setLevel(log_level)  # and reset
 
+    # Now if the input arrays where splitted in chunks, we need to gather
+    # the data arrays for the additional attributes.
     other = result.meta['other']
     if other and len(boxes) > 1:
         for (name, coords), obj in list(other.items()):
--- a/astrodata/fits.py
+++ b/astrodata/fits.py
@@ -1577,7 +1577,7 @@ def windowedOp(func, sequence, kernel, shape=None, dtype=None,
                                  .format(shape, kernel))
         ticks = [[(x, x+step) for x in range(0, axis, step)]
                  for axis, step in zip(shape, kernel)]
-        return cart_product(*ticks)
+        return list(cart_product(*ticks))
 
     if shape is None:
         if len({x.shape for x in sequence}) > 1:
@@ -1602,19 +1602,39 @@ def windowedOp(func, sequence, kernel, shape=None, dtype=None,
     # The Astropy logger's "INFO" messages aren't warnings, so have to fudge
     log_level = astropy.logger.conf.log_level
     astropy.log.setLevel(astropy.logger.WARNING)
+
+    boxes = generate_boxes(shape, kernel)
+
     try:
-        for coords in generate_boxes(shape, kernel):
-            # The coordinates come as:
-            # ((x1, x2, ..., xn), (y1, y2, ..., yn), ...)
-            # Zipping them will get us a more desirable:
-            # ((x1, y1, ...), (x2, y2, ...), ..., (xn, yn, ...))
-            # box = list(zip(*coords))
+        for coords in boxes:
             section = tuple([slice(start, end) for (start, end) in coords])
-            inputs = [element.window[section] for element in sequence]
-            result.set_section(section, func(inputs, **kwargs))
+            out = func([element.window[section] for element in sequence],
+                       **kwargs)
+            result.set_section(section, out)
+
+            # propagate additional attributes
+            if out.meta.get('other'):
+                for k, v in out.meta['other'].items():
+                    if len(boxes) > 1:
+                        result.meta['other'][k, coords] = v
+                    else:
+                        result.meta['other'][k] = v
+
             gc.collect()
     finally:
         astropy.log.setLevel(log_level)  # and reset
+
+    other = result.meta['other']
+    if other and len(boxes) > 1:
+        for (name, coords), obj in list(other.items()):
+            if not isinstance(obj, NDData):
+                raise ValueError('only NDData objects are handled here')
+            if name not in other:
+                other[name] = NDDataObject(np.empty(shape,
+                                                    dtype=obj.data.dtype))
+            section = tuple([slice(start, end) for (start, end) in coords])
+            other[name].set_section(section, obj)
+            del other[name, coords]
 
     return result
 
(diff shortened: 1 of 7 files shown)


astrodata/fits.py
```diff
@@ -1578,14 +1602,49 @@ def generate_boxes(shape, kernel):
...
+                section = tuple([slice(start, end) for (start, end) in coords])
+                other[name].set_section(section, obj)
+                del other[name, coords]
+
+        for name in other:
+            # To set the name of our object we need to save it as an ndarray,
+            # otherwise for a NDData one AstroData would use the name of the
+            # AstroData object.
+            other[name] = other[name].data
+            result.meta['other_header'][name] = fits.Header({'EXTNAME': name})
```
saimn, line 1647: So the most tricky part is here, because `windowedOp` can split the NDData objects in blocks. So in the loop I save the additional NDData objects (in `.meta['other']`) with their coordinates, and then I recreate the final object. But this way has some limitations, like I cannot guess the shape so I use the input one, but because of that I cannot save the full cube of the individual masks. I guess saving the "rejection map", i.e. the sum over the num_img axis, is good enough ?

Also another issue here: if I save a NDData object, AstroData has no way to set its extension name... But if I save a ndarray I can set its EXTNAME with the header:
https://github.com/GeminiDRSoftware/DRAGONS/blob/c952cd99d5a39a384b35d3ae9b64812bdf80bd2b/astrodata/fits.py#L1047-L1053

geminidr/core/primitives_stack.py
```diff
@@ -170,19 +171,12 @@ def stackFrames(self, adinputs=None, **params):
...
-                    except AttributeError:  # For non-lazy VAR
-                        bytes += item._array.dtype.itemsize
-            bytes += 2  #  mask always created
-            bytes_per_ext.append(bytes * np.multiply.reduce(ext.nddata.shape))
+            bytes += 2 * ext.data.dtype.itemsize
+            if ext.variance is not None:
+                bytes += ext.variance.dtype.itemsize
+
+            bytes += 2  # mask always created
+            bytes_per_ext.append(bytes * np.product(ext.shape))
```
saimn, line 179: While looking at memory usage computation I found that 1) the final NDData object is not taken into account, so I cannot add the additional array here, and 2) it could be greatly simplified. Hope it's fine?

gempy/library/nddops.py
```diff
@@ -44,35 +44,6 @@
...
-    inds = []
-    for dim, n in zip(dest_dims, arr.shape):
-        if dim is None:
-            inds.append(ind)
-        else:
-            ind_shape_dim = ind_shape[:dim] + (-1,) + ind_shape[dim+1:]
-            inds.append(np.arange(n).reshape(ind_shape_dim))
-    return arr[tuple(inds)]
-
-
```
saimn, line 75: Removed this, since I also need `np.put_along_axis` and Astropy 4.0 requires Numpy 1.16 anyway. So I replaced this function below with `np.take_along_axis`, but interestingly the Numpy version does not expand dimensions when needed, so I needed to handle this with `np.expand_dims`.

gempy/library/nddops.py
```diff
@@ -178,24 +155,26 @@ def __init__(self, combine='mean', reject='none', log=None, **kwargs):
...
-            self._logmsg('No such rejector as {}. Using none instead.'.format(reject),
-                         level='warning')
+            self._logmsg('No such rejector as {}. Using none instead.'
+                         .format(reject), level='warning')
             rejector = self.none
+
         req_args = getattr(rejector, 'required_args', [])
         self._rejector = rejector
-        self._dict.update({k: v for k, v in kwargs.items() if k in req_args})
+        self._rej_args = {k: v for k, v in kwargs.items() if k in req_args}
```
saimn, line 177: Instead of merging combine and rejection args and then splitting them when they are used, I added two variables.

gempy/library/nddops.py
```diff
@@ -283,47 +262,55 @@ def _num_good(mask):
...
-                     if arg in self._dict}
 
         if self._debug_pixel is not None:
-            self._pixel_debugger(data, mask, variance, stage='after rejection')
-            self._logmsg("Combining: {} {}".format(self._combiner.__name__, comb_args))
+            self._logmsg("Debug pixel coords {}".format(self._debug_pixel))
+            self._pixel_debugger(data, mask, variance, stage='at input')
+            info = data[(slice(None),) + self._debug_pixel]
+            self._logmsg("stats: mean={:.4f}, median={:.4f}, std={:.4f}"
+                         .format(np.mean(info), np.median(info), np.std(info)))
```
saimn, line 287: Added a few info here, the mean/median/std at the beginning of the clipping, which can help to understand the clipping process.

gempy/library/nddops.py
```diff
@@ -486,28 +478,36 @@ def minmax(data, mask=None, variance=None, nlow=0, nhigh=0):
...
+                    np.put_along_axis(newmask, idx, ONE, axis=0)
+
+                # high values
+                idx = np.expand_dims(arg[i][(i > nhi) & (i < num_good)], axis=0)
+                if idx.size > 0:
+                    np.put_along_axis(newmask, idx, ONE, axis=0)
+
+            mask |= newmask
+            # mask[i][i < nlo] |= ONE
+            # mask[i][np.logical_and(i > nhi, i < num_good)] |= ONE
```
saimn, line 510: minmax was sorting the data and mask, which would have been an issue to compute the rejection mask. So I updated this, the case without mask was easy with `np.put_along_axis`, but the other case is more tricky ;)

geminidr/core/primitives_stack.py
chris-simpson: With each cycle of `windowedOp`, are the output arrays stored directly in the final array or do the get created elsewhere in memory and then get copied? If it's the first, then this memory should *not* be included in the calculation. The memory limit provided is purely the space that can be used for calculations during each `windowedOp` cycle and is then freed up again at the end.

chris-simpson:
I'm not going to have time to look at this in too much detail. Note that there is a branch `fix_nddops` which has a code update in `cyclip.pyx` and then a subsequent tweak in `nddops.py`. This was designed to fix the problem where no rejection took place when all pixels were bad (so, for example, cosmic ray hits weren't rejected in the unilluminated border of GMOS images). Can I suggest you write some better tests for `NDStacker` than what we have now and merge the `fix_nddops` branch into yours and then merge that into `master` after confirming the behaviour?

geminidr/core/primitives_stack.py
saimn: The stacked array is stored in memory and then is set (with `set_section`) in the final output one, so at some point there is an additional array in memory.

saimn:
Hmm ok, the minmax part will clearly conflict with what I did, but I can try to merge both versions while this is still fresh.

PR #64, opened 2020-05-22
refac/store processed science

olyoberdorf:
per Kathleen and Chris, renaming as storeProcessedScience and updating references

also now ignoring 'qa' mode files in the method


chris-simpson:
I looked at the diff, seems fine.

PR #65, opened 2020-05-22
Added flux calibrate to LS_SPECT

olyoberdorf:
This adds the fluxCalibrate step to the quicklook LS_SPECT recipe


olyoberdorf:
forgot this is subsumed by other pr

PR #66, opened 2020-05-28
Refactor Astrodata

saimn:
I think this branch is in a good enough state for a merge, as a first step.
~~(And it is waiting for the Gwcs branch with which I will need to fix some conflicts.)~~

What is done here:
- essentially merging `DataProvider`, `FitsProvider` and `FitsProviderProxy` in the `AstroData` class. (Better to have a look directly at `astrodata/core.py` given the number of changes).
- split the reading and writing of FITS files to dedicated functions in the `fits` module, but those are still used directly for `astrodata.open` and `ad.write`. 
- no breaking change!

What remains to be done:
- removing `AstroDataFits`. Currently this is an empty class since it is inherit by other classes in DRAGONS. This is just a matter of changing the inherited class, but I will keep that for a following PR to limit the number of changes here.
- I think the main item is how to replace `extver` with something more generic, e.g. a list of indices. Need some thoughts on the best way to do this.
- Find a better solution for associated objects. Currently these are stored directly in `__dict__`, with a special treatment for tables (with `._tables`). I think it would be cleaner with a dedicated dict, e.g. `_other`.
- Replace `AstroDataFitsDeprecationWarning` with `AstroDataDeprecationWarning` ? If `AstroDataFits` is gone, the latter makes more sense.

cc @chris-simpson @jehturner

### DIFF
--- a/astrodata/__init__.py
+++ b/astrodata/__init__.py
@@ -7,13 +7,13 @@ accordingly to the :class:`~astrodata.TagSet` received.
 
 __all__ = ['AstroData', 'AstroDataError', 'TagSet', 'NDAstroData',
            'astro_data_descriptor', 'astro_data_tag', 'keyword',
-           'open', 'create', '__version__', 'version']
+           'open', 'create', '__version__', 'version', 'add_header_to_table']
 
 
-from .core import AstroData, AstroDataError, AstroDataFits
+from .core import AstroData, AstroDataFits
 # TODO: Remove 'write' when there's nothing else using it
 from .fits import KeywordCallableWrapper, add_header_to_table
-from .factory import AstroDataFactory
+from .factory import AstroDataFactory, AstroDataError
 from .nddata import NDAstroData
 from .utils import *
 from ._version import version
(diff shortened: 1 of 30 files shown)


chris-simpson:
Re: EXTVER. This is currently stored in `ext.nddata.meta['ver']`, which I assume will be accessible via `ext.meta['ver']` based on the underlying concept of passing unrecognized attributes to the `NDAstroData` object for processing. And there's an `extver_map()` which tells you the index corresponding to a given EXTVER (I'm not sure this is ever used). Most of the accessing of EXTVER is for logging, where we use this rather than the index because users familiar with Gemini-IRAF think `[0]` is the PHU. And when opening a FITS file, the slices are ordered by EXTVER. But do we want to preserve this? If a file is read with two extensions, EXTVER=1 and 4, for example, should those values remain when the file is written back to disk, or should they become 1 and 2, because only the _order_ matters (which is stored internally because the AD slices are an ordered list)? Even if that's the case, I think it would still be nice to have something like `ext.ver` that refers to the index of this slice so that useful information can be logged in a `for ext in ad` loop instead of having to do `for extver, ext in enumerate(ad)`.

saimn:
All tests pass, so I guess it's time to merge this :crossed_fingers:

PR #68, opened 2020-06-17
Updated SkyCam AstroData implementation to latest underlying classes

olyoberdorf:
The SkyCam class had fallen behind some of the updated internals of AstroData.  I've updated it to work off the phu/as appropriate.

TBD: I can add unit tests if someone walks me through how to host the datafile and best practices for DRAGONS


olyoberdorf:
Another thought I had - maybe we should consider ways to have DRAGONS-as-a-library users register their own AstroData types into the factory.  If SkyCam logic isn't really something that belongs in DRAGONS (and I have no opinion either way) then it could be something defined in the FITSStore and we "register" it.

Sort of O.T. but it struck me while adding this change.

chris-simpson:
I have never seen any SkyCam data and have no way to know whether any of this is right or not.

gemini_instruments/skycam/adclass.py
```diff
@@ -36,34 +37,23 @@ def _tag_site(self):
...
     def dec(self):
-        hdr = self.TABLE1.meta.get('header')
-        return hdr.get('DEC')
-    
-    @astro_data_descriptor
-    def ut_datetime(self):
-        hdr = self.TABLE1.meta.get('header')
-        return dateutil.parser.parse(hdr.get('DATE-OBS'))
-        
+        return self.phu['DEC']
```
saimn, line 56: Then there is probably no need to redefine those descriptors (ra, dec, exptime, ...) since the parent class is already getting those values from the primary header ?

olyoberdorf: The code in the base class actually looks a little different to this.  For RA and DEC, if I remove these overrides I no longer get values for the skycam files.  If you want a sample file to play with let me know.

saimn: Yes, out of curiosity ;)

chris-simpson: I find this surprising, since the ra() descriptor tries to get the location from the WCS of the first extension -- although the _get_wcs_coords() method should probably convert (CRPIX1-1, CRPIX2-1) instead of the middle of the extension -- and then falls back to the phu['RA'] keyword (and dec() does the same), so I would have expected this to work.

olyoberdorf: I was able to remove exposure_time and ut_time and the files still ingest for archive.  I've left RA and DEC since it seems to not like something about the wcs bit in the base class.  I think that's probably fine for these files, but if someone wants to work through that with me let me know.

olyoberdorf: The WCS call whines about "invalid projection" somewhere deep in the internals and skips over the rest by throwing an exception.  It's a ValueError with this notation in it:

`ERROR 5 in wcsset() at line 2279 of file cextern/wcslib/C/wcs.c:
Invalid parameter value.
ERROR 2 in celset() at line 215 of file cextern/wcslib/C/cel.c:
Invalid projection parameters.
ERROR 2 in zpnset() at line 2213 of file cextern/wcslib/C/prj.c:
Invalid parameters for zenithal/azimuthal polynomial projection.
`

chris-simpson: Yeah, what the heck is "RA---ZPN"? Google tells me it's a "Zenith Polynomial Projection" but it's missing polynomial keywords. Perhaps it means "RA---AZP"? for Zenithal Projection?

AD-with-gWCS handles that and produces something. And, tbh, the descriptors should now be using the wcs attributes, not reconstructing them from the FITS headers, so those descriptors will need to be refactored anyway.

kiloRomeoAlpha:
"If SkyCam logic isn't really something that belongs in DRAGONS ..."

I'll just note here that this was done way back when the new Astrodata made it incredibly easy to accommodate new instruments. When the idea of archiving 'skycam' data was presented, the AstroDataSkyCam class was done up as a quick demo. (on the now defunct sbffits-dev-lv1) as proof-of-concept. And, indeed, the skycam data then was being ingested by the dev server. And then it sat there for ... years.

These days, I'm not convinced that AstroDataSkyCam belongs in DRAGONS. I once thought about having an AstroDataGOA (subclassed on Gemini) or AstroDataFitsstore or something that the fitsstores and GOA would use and modify for other non-DRAGONSy instruments without need to modify the DRAGONS package. You will note many instrument packages are defined in gemini_instruments, almost none of which will ever be used except by the Fitsstores. Of course, it is nice that Astrodata can handle all the legacy instruments, though I am not sure how useful it would been seen by the outside world.

olyoberdorf:
Yeah that sounds exactly like what I was thinking.  Certainly would allow for moving any data hacks as far away from our core libraries as possible.  Obviously better if that's done on the datafile side, but if not perhaps it's better to have it in GOA than in DRAGONS.

phirstgemini:
I'm not so up to speed on what counts as DRAGONS or "DRAGONS core", but the
whole point of AstroData was a data abstraction layer to deal with and hide
away the quirks of data from individual quirky instruments. We shouldn't be
moving that functionality into fits storage or the archive code for
example. I think a user would rightfully be very frustrated if they
download gemini data from the gemini archive and find it won't open with
astrodata that they download from gemini.

chris-simpson:
The WCS in the header is wrong so, for example, ds9 throws out errors all the time. We have two options:

1) Write bespoke ra() and dec() descriptors that just use the PHU keywords. This will allow the archive to ingest it but the WCS is crap so there's a limited amount anybody could do with the data (e.g., they couldn't look at an image and work out where in RA and DEC a transient had happened).

2) Edit the headers to RA---ZPN reads RA---AZP (and same for DEC), and get future headers to write this. Then the Gemini-level ra() and dec() descriptors work fine.
2a) Also update the Gemini-level descriptors to return the RA and DEC of (CRPIX1, CRPIX2) rather than the center of the image.

(1) is fine if we're just ingesting the data but nobody is ever going to want to look at it. But then why are we ingesting it? So I recommend (2).

kiloRomeoAlpha:
_"a user would rightfully be very frustrated if they download gemini data from the gemini archive and find it won't open with astrodata that they download from gemini."_

Yes, this is the obvious good point. But let us note that the SkyCam is not actually Gemini data per se. It is different "telescope" and why subclassing it on AstroDataGemini is not straight forward.

So, in the interests of SUSD and DRAGONS reputation, I agree with Chris that if we are going to have the skycam data in the archive, it should be in properly. It will also be true that no one, or almost no one, will ever look at this.

KathleenLabrie:
Just a bit of "top-level" view here.   "gemini_instruments" is just one astrodata support package.   The idea is that there can be multiple "something_instruments" that can be imported as needed.  We considered the idea before when we had a flurry of visitor instruments and were not sure how to deal with them.  I'd argue that there are instruments in "gemini_instruments" that don't really belong there.  (BTW, there is already a "skycam" in there.)   

Some review of all this will need to be done at some point, though not now as we have other priorities.  As was pointed out, right now, the system being new and our users being unfamiliar with it, having everything in "gemini_instruments" for data that can be downloaded from the archive is probably the best compromise.

As for the value of skycam in the main archive, someone messed up back then by entertaining this stupid idea and now we are stuck with it.

PR #72, opened 2020-07-23
Allow passing a recipe directly as a callable

saimn:
@kiloRomeoAlpha - This is the change I proposed some time ago, to allow passing directly a recipe function. What do you think about it ?

### DIFF
--- a/recipe_system/mappers/recipeMapper.py
+++ b/recipe_system/mappers/recipeMapper.py
@@ -32,7 +32,11 @@ class RecipeMapper(Mapper):
 
     """
     def get_applicable_recipe(self):
-        recipefn = find_user_recipe(self.recipename)
+        if callable(self.recipename):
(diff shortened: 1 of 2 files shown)


KathleenLabrie (commented):
@chris-simpson, @saimn What to we think about this old pull request from Ken?

chris-simpson:
Isn't it a PR from Simon? So presumably Simon likes it. I've never had cause to use the `Reduce()` API (I've only come across it in `autoredux` when sorting out the SExtractor memory issues) so I can't comment except saying it looks OK and it seems like it adds useful functionality.

KathleenLabrie:
Yes!  Sorry.   I only saw the bold kiloRomeoAlpha it was addressed to.

If we all agree then, let's do it.

saimn:
I also forgot about this PR ;). I guess I can merge it then.

PR #74, opened 2020-08-13
Enable travis again for DRAGONS

b1quint:
I don't remember why we stopped using Travis in the past but it is now running again. I merged `master` into this branch just now and all the tests are passing. 

Again, the idea is that Travis only runs Unit Tests that do not require: 1) pre-processed data, 2) reference data. If this works and people get used to it, we can start splitting the load of the tests between Jenkins and Travis. Another option is to add a marker telling whether the tests should only run in Jenkins or Travis (or both).


.travis.yml
```diff
@@ -0,0 +1,32 @@
+os: linux
+dist: xenial
+language: python
+python:
+  # We don't actually use the Travis Python, but this keeps it organized.
+  - "3.7"
+env:
+  global:
+    - DRAGONS_TEST="$HOME/dragons_tests/"
+    - JENKINS_CONDA_HOME="$HOME/miniconda"
```
saimn, line 10: Not needed ?

.travis.yml
```diff
@@ -0,0 +1,32 @@
...
+  - conda config --set restore_free_channel true
+  - conda env list
+  - conda install --yes pip wheel
+  - pip install "tox<=3.17.1" tox-conda
+
+  # Useful for debugging any issues with conda
+  - conda info -a
+
+script:
+  - tox -e py37-unit -v -- --junit-xml reports/unittests_results.xml
```
saimn, line 32: `--junit-xml` is useful only for jenkins I think ?

.travis.yml
b1quint: `DRAGONS_TEST` is the root folder where the files downloaded by the `download_from_archive` function will be cached. Removed `JENKINS_CONDA_HOME`.

PR #75, opened 2020-08-17
Add fit_1D function

saimn:


jehturner:
Thanks, that's very helpful! I still have a couple more changes to commit to this, but I can do that in `master` after it's merged. Then I can probably just reset my branch to before the last 3 commits and rebase it.

saimn:
Ok, merging then :)

jehturner:
I think this looks good, thanks. One pedantic detail is that your second commit message just says "Add option for debug plot", but the most important change in that commit is actually renaming `iters` -> `maxiters`.

I wasn't 100% sure what you intended with the plotting (my original intention wasn't that clear either), but I think it's important for the masked points to be  plotted at their actual y values, with a different symbol or ideally colour from the unmasked ones (I think red is intuitive for points that are considered "bad"). But of course the plotting is just provisional anyway.

jehturner:
I think best just leave the commit message as it is now, since changing it would involve a messy rebase? There's always the diff if someone really wants to see what changed...

saimn:
Yes, now that the commit is on master it is better not to touch it...
For the plotting I think I just un-commented the existing code to have a look, and thought it would be useful later.

jehturner:
Yes, it's useful for debugging and also to sketch out what needs plotting & make sure that information is available in the right form; I was just adding annotation of user-specified regions.

PR #76, opened 2020-08-19
Fix/test clean up

b1quint:
Refactored integration tests to clean up output files. A `--keep-data` option was added in order to keep pre-stack data. By default, all the pre-stack data is deleted once Reduce.runr() is done.


b1quint:
[Test clean-up](https://trello.com/c/C8XqEjoo/1-test-clean-up)

PR #78, opened 2020-08-29
Updated to fallback to RA, DEC header keywords for unusable WCS 

olyoberdorf:
This is the same logic currently seen in the core Gemini support, just applying it to GNIRS' custom AstroData logic for ra/dec

### DIFF
--- a/gemini_instruments/gnirs/adclass.py
+++ b/gemini_instruments/gnirs/adclass.py
@@ -431,6 +431,8 @@ class AstroDataGnirs(AstroDataGemini):
         # if it's messed up
 
         wcs_ra = self.wcs_ra()
+        if wcs_ra is None:
+            return self.phu.get('RA', None)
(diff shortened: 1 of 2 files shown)


b1quint:
[Review PR #78](https://trello.com/c/zuR0xVXx/2-review-pr-78)

b1quint:
Hi Oliver,

I created a test that checks if `ad.ra()` and `ad.dec()` will always return a float. In my test case, the input file was fine and `ad.wcs_ra()`/`ad.wcs_dec()` returned floats. I am wondering if you could add the file you found in which `ad.wcs_*()` returned None to the test cases.

b1quint (approved):
I created a test to verify the changes but it has only one test file. I believe we can merge as soon as we add more test files showing the different cases.

b1quint (approved):

PR #84, opened 2020-09-25
Some fixes for UnivariateSplineWithOutlierRemoval

saimn:
While looking at the skysub branch with James, and at reusing the rejection mask in the `UnivariateSplineWithOutlierRemoval` iterations for the sigma-clipping, I found a few issues:
- the mask passed to `sigclip` was ignored, because since 588b011 and be84d5d the values passed to sigclip are ignored only if their mask value is < 32768
- masked pixels were grown only to the right.

### DIFF
--- a/geminidr/core/primitives_spect.py
+++ b/geminidr/core/primitives_spect.py
@@ -1900,15 +1900,11 @@ class Spect(PrimitivesBASE):
         log = self.log
         log.debug(gt.log_message("primitive", self.myself(), "starting"))
         timestamp_key = self.timestamp_keys[self.myself()]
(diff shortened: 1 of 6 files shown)


jehturner (commented):
I think this looks good, as long as the masking remains cumulative, ie. no points get unmasked in the sigma clipping (otherwise the subsequent XOR isn't going to work, which seems like a possible future regression). Speaking of which, are we going to add a couple of tests for these fixes? I think it would be OK to say that we'll cover them with the `fit_1D` tests we were already planning, as long as the corrections get covered somehow.

saimn:
Merged into `sky_sub`.

PR #86, opened 2020-09-29
Use malloc/free to allocate tmp arrays in cyclip

saimn:
This is to avoid segfaults if using `iterclip` with an array with more than 10'000 elements. 
Also for the case of stacking a few images it will avoid allocating unnecessarily big arrays.

### DIFF
--- a/gempy/library/cyclip.pyx
+++ b/gempy/library/cyclip.pyx
@@ -16,8 +16,9 @@ this command again under the new environment.::
 """
 
 import numpy as np
-from libc.math cimport sqrt
 cimport cython
+from libc.math cimport sqrt
+from libc.stdlib cimport malloc, free
 
 
 @cython.boundscheck(False)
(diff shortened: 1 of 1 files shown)


chris-simpson:
That looks reasonable, from what I remember of C.

phirstgemini (commented):
You really should to check the return value of malloc - it will return NULL if it's unable to allocate the memory. Sure, that shouldn't normally happen, but you'll get a segfault if it does and I'm sure you'll want to raise something rather than let that happen..

Sorry if I'm wrong and cython takes care of this for you, but it didn't last time I worked with it...

saimn:
Yes, you're right. In the `def` function we can raise a `MemoryError`, I need to find the good way for the `cdef` function.

phirstgemini:
I think you can just do
```
if not tmp:
  raise MemoryError()
```

immediately after the malloc line, no?

saimn:
The first occurrence is in a `cdef` function, which cannot handle exceptions, so you define a return value that will be checked if an exception happen: https://cython.readthedocs.io/en/latest/src/userguide/language_basics.html#error-return-values

phirstgemini:
Ah yes - good catch, I'd forgotten about that!

PR #94, opened 2020-11-03
Setup github actions

saimn:


b1quint:
Quick question: do you know how much disk space we have available on GitHub Actions? 

Unrelated question: I think I saw Github Action being called as a cron job but I assume it also runs on every push, right?

b1quint (approved):

saimn:
About the cache, "total size of all caches in a repository is limited to 5 GB." (https://docs.github.com/en/free-pro-team@latest/actions/guides/caching-dependencies-to-speed-up-workflows#usage-limits-and-eviction-policy)
I started a new build which uses the cache, and currently it uses a bit less than 1Gb.

Yes, it runs on every push, and will run weekly with the cron. I removed the run on pull_request event because we all use branch on the repo (not on a fork), so to avoid running it twice when making a PR. And it should also be possible to run it manually if needed.

PR #103, opened 2020-11-27
Re-factor fit1d as a class and make a few requested improvements

jehturner:
Here's a PR for my refactoring so you can see a summary of the changes (though they're not that easy to read since there have been changes of indentation etc., but I did most of that in a separate commit). I still have several "to do" improvements, as I keep thinking of more issues, but this should be good enough for now. I might just merge this into `sky_sub` (which should itself be ready to merge into `master` before long) by Monday whether or not there are comments, but it's here to summarize the changes and for any discussion.

I miss the IRAF parameter names a bit, as I don't have to think about them. Never mind...

### DIFF
--- a/geminidr/core/primitives_spect.py
+++ b/geminidr/core/primitives_spect.py
@@ -2350,10 +2350,11 @@ class Spect(PrimitivesBASE):
                 # but should we include some specific set of DQ codes here?
                 sky = np.ma.masked_array(ext.data, mask=sky_mask)
 
-                sky_model = fit_1D(sky, weights=sky_weights, function=function,
-                                   order=order, axis=axis, sigma_lower=lsigma,
-                                   sigma_upper=hsigma, niter=max_iters,
-                                   grow=grow, regions=regions, plot=debug)()
+                sky_model = fit_1D(
+                    sky, weights=sky_weights, function=function, order=order,
+                    axis=axis, sigma_lower=lsigma, sigma_upper=hsigma,
+                    niter=max_iters, grow=grow, regions=regions, plot=debug
+                ).evaluate()
 
                 ext.data -= sky_model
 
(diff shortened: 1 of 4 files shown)


saimn:
Looks good to me, the only remark I have is that when reading code like
```
        fit = fit_1D(self.data, weights=self.weights, function='chebyshev',
                     order=1, axis=0, sigma_lower=2.5, sigma_upper=2.5,
                     niter=5, plot=debug)
        fit_vals = fit()
```
I would imagine that the fit is done by `fit()`. So maybe this should be an explicit method (`fit.evaluate()` ?) instead of using `__call__` ?

jehturner:
Yes, I'm not sure about changing it to a method, but I called it `fit` rather hurriedly and with some doubt as to whether that was really the best name here (any better suggestions?). We could vote on it but Chris seems to like `__call__`. @olyoberdorf.

We could always call it `fits` so it's clear it's a noun :grin:.

Any idea why that codecov job is bizarrely complaining about GSAOI?

saimn:
I think codecov is troubled by the reports sent both from Github Actions and Jenkins (which sends reports for the different stages, though Actions run only unit tests). I removed the codecov reporting from Actions, on master, so it will wait for Jenkins to run before reporting coverage.

jehturner:
Maybe `fit1d  = fit_1D()` (or even `fitted`, though I find an orphaned adjective a bit distasteful)? There would be some logic in reversing the terms, but in this case it (`1dfit`) doesn't work for multiple reasons.

jehturner:
The mixture (mess?) of Python & IRAF conventions that we've ended up with here bothers me a bit, but it's pretty much an unavoidable side effect of wanting to present Cartesian co-ordinates to users. The rule of thumb is that if a variable refers to pixels along the axis being fitted it's 1 indexed and inclusive, otherwise it follows Python conventions. That's as self-consistent as I could really make it without other compromises.

Along those lines, I should perhaps change the `points` argument of `__call__` to take 1-indexed pixels, given that `regions` does (though it can now also take Python slices) and that's what the built-in plotting shows? I previously thought of some complication with changing that but it escapes me now.

jehturner:
Decided on Slack to change `__call__()` to `evaluate()`.

jehturner:
I'm going to merge this so we can finalize `sky_sub`, but that shouldn't stop you adding comments if you want. We could probably improve the codecov stats a bit when there's time, but this already has more tests than it had before.

Issue #104, opened 2020-12-02
Error in GMOSSpect().findSourceApertures() -- TypeError: Improper input: N=3 must not exceed M=0


b1quint:
Tests for `p.findSourceApertures()` are now failing. The traceback is attached.
[traceback.txt](https://github.com/GeminiDRSoftware/DRAGONS/files/5629201/traceback.txt)

PR #105, opened 2020-12-02
Fix test_find_apertures_with_fake_data()

b1quint:
Solves Issue #104. The problem was not with the primitive itself, but with the test. 

It was caused because I was trying to create a variance plane using the square root of the data, which contains negative values and created NaNs.


b1quint:
The tests that are failing are not related to this PR. Merging now.

PR #92, opened 2020-10-29
Simplify setup and declare dependencies

saimn:

PR #106, opened 2020-12-03
Run tests with the installed code

saimn:
- Run tests with the installed code, to detect packaging issues. This is done with tox.
- To achieve that, I moved our pytest "plugin" from `conftest.py` to a real plugin (`pytest_dragons`).
- This branch is based on #92 
- TODO: find another way to get the version, instead of importing it from astrodata.


KathleenLabrie:
Do you need input from the team to complete this one?

saimn:
I got some feedback from @b1quint who wanted to keep the possibility to run tests for a specific file, which seems reasonable. But while trying to find a solution I found a problem with tox-conda (https://github.com/tox-dev/tox-conda/issues/64) and then moved to something else. The tox-conda issue is a blocker right now, though I could extract some useful bits from this PR like moving the pytest specific code to a plugin.

saimn:
I think it's good to go now, so merging.
Note: with the latest changes it's possible to run tox for a specific file but with the module name instead of a path, e.g. `tox -v -e py37-unit -- astrodata.tests.test_fits`.

PR #107, opened 2020-12-18
Add fixPixels primitive

saimn:


geminidr/core/parameters_preprocess.py
```diff
@@ -53,6 +53,20 @@ class dilateObjectMaskConfig(config.Config):
...
     repeat = config.Field("Allow dilation of already-dilated image?", bool, False)
 
+
+class fixPixelsConfig(config.Config):
+    suffix = config.Field("Filename suffix", dtype=str,
+                          default="_pixelsFixed", optional=True)
+    regions = config.Field("Regions to fix", dtype=str, optional=True)
+    regions_file = config.Field("Path to a file containing the regions to fix",
+                                dtype=str, optional=True)
+    axis = config.Field("Axis over which the interpolation is done",
```
KathleenLabrie, line 63: Shouldn't there be a default defined for `axis`?  None or zero for automatic mode, or something?

geminidr/core/parameters_preprocess.py
```diff
@@ -53,6 +53,20 @@ class dilateObjectMaskConfig(config.Config):
     dilation = config.RangeField("Dilation radius (pixels)", float, 1., min=0)
     repeat = config.Field("Allow dilation of already-dilated image?", bool, False)
 
+
+class fixPixelsConfig(config.Config):
+    suffix = config.Field("Filename suffix", dtype=str,
+                          default="_pixelsFixed", optional=True)
+    regions = config.Field("Regions to fix", dtype=str, optional=True)
```
KathleenLabrie, line 60: I think that it would be good to have an example of the format expected for the string defining regions to fix.

geminidr/core/primitives_preprocess.py
```diff
@@ -533,6 +501,143 @@ def dilateObjectMask(self, adinputs=None, suffix=None, dilation=1, repeat=False)
...
         return adinputs
 
+    def fixPixels(self, adinputs=None, suffix=None, axis=None, regions=None,
+                  regions_file=None, use_local_median=False, debug=False):
+        """
+        This primitive replaces bad pixels by linear interpolation along
+        lines or columns using the nearest good pixels, similar to IRAF's
+        fixpix.
+
+        Regions must be specified either as a string, separated by semi-colons,
```
KathleenLabrie, line 511: An example would be helpful.

KathleenLabrie (commented):
I've mostly focused on the user interface and added a couple comments in the file diffs.

@chris-simpson is better suited to comment on the implementation.

geminidr/core/parameters_preprocess.py
saimn: The default is None but there is no need to specify it here, I think it's implicit with `optional=True`. 
`showpars` gives this:
```
 Name			Current setting

suffix               '_pixelsFixed'       Filename suffix
regions              None                 Regions to fix
regions_file         None                 Path to a file containing the regions to fix
axis                 None                 Axis over which the interpolation is done
use_local_median     False                Use a local median filter for single pixels?
```
0 would be the first axis so it cannot be the value for the automatic mode.

geminidr/core/parameters_preprocess.py
saimn: I'm not sure if putting a longer multi-line string here is a good idea or not ? It seems to be managed correctly by `showpars` does it look good to you ?
```
========================================
 Name			Current setting

suffix               '_pixelsFixed'       Filename suffix
regions              None                 Regions to fix.

Regions must be specified either as a string, separated by semi-colons,
with the ``regions`` parameter, or with a file (``regions_file``), one
region per line.

Regions strings must be a comma-separated list of colon-separated pixel
ranges, one per axis, in 1-indexed Cartesian pixel co-ordinates,
inclusive of the upper limit. Axes are specified in Fortran order
(reverse of the Python order). For example, "450,521;430:437,513:533"
would specify a single pixel (line 521, column 450) and a vertical
region (lines 513 to 533 included, columns 430 to 437).
    
regions_file         None                 Path to a file containing the regions to fix
axis                 None                 Axis over which the interpolation is done
use_local_median     False                Use a local median filter for single pixels?
```

chris-simpson:
I'm going to comment in more than one reply, so we can use github's ability for multiple conversations, particularly as this first one is likely to prompt discussion.

First, I'm not a fan of the long Config `doc` string, because it's inconsistent with all the other strings. There's a large discussion here about what `showpars` should do compared to the more extensive docstrings in the primitives (and how to access those docstrings from the shell). My thought was always that `showpars` would behave like IRAF's `lpar` and provide a list of parameters with very short descriptions (just aides-memoires) whereas detailed help, more analogous to IRAF's `help`, (which would probably be the primitive's docstring) would appear somewhere else (I think I suggested `smaug` as the command, for SMArt User Guide). My main argument against having long strings in `showpars` is that the user almost certainly will want the long explanatory text from `smaug` if they're trying to learn the primitive for the first time and we obviously don't want to duplicate the text.

geminidr/core/parameters_preprocess.py
chris-simpson: I'm not a fan of the long Config doc string, because it's inconsistent with all the other strings. There's a large discussion here about what showpars should do compared to the more extensive docstrings in the primitives (and how to access those docstrings from the shell). My thought was always that showpars would behave like IRAF's lpar and provide a list of parameters with very short descriptions (just aides-memoires) whereas detailed help, more analogous to IRAF's help, (which would probably be the primitive's docstring) would appear somewhere else (I think I suggested smaug as the command, for SMArt User Guide). My main argument against having long strings in showpars is that the user almost certainly will want the long explanatory text from smaug if they're trying to learn the primitive for the first time and we obviously don't want to duplicate the text.

I also don't see the need for separate `regions` and `regions_file` parameters. Generally I like to keep the number of parameters small, perhaps to an unreasonable extent, but I think a single parameter works here.

geminidr/core/primitives_preprocess.py
```diff
@@ -533,6 +501,147 @@ def dilateObjectMask(self, adinputs=None, suffix=None, dilation=1, repeat=False)
...
+                            plot_slices = [Ellipsis] + plot_slices[-2:]
+                        origdata = ext.data[plot_slices].copy()
+
+                    if use_local_median and np.prod(region_shape) == 1:
+                        nd = NDAstroData(data=ext.data,
+                                         mask=np.zeros(ext.shape, dtype=bool))
+                        nd.mask[slices] = 1
+                        ring_median_filter(nd, 3, 5, inplace=True,
+                                           replace_flags=1,
+                                           replace_func='median')
```
chris-simpson, line 603: Instead of creating a new (possibly large) `mask` array just to flag one pixel, would it be better to set a high bit of the "bad" pixel and pass that to the function?
```
ext.mask[slices] |= 32768
ring_median_filter(..., replace_flags=32768, ...)
ext.mask[slices] ^= 32768
```

geminidr/core/primitives_preprocess.py
```diff
@@ -533,6 +501,147 @@ def dilateObjectMask(self, adinputs=None, suffix=None, dilation=1, repeat=False)
...
+
+        for ad in adinputs:
+            for ext in ad:
+                for region in regions:
+                    try:
+                        slices = cartesian_regions_to_slices(region.strip())
+                    except ValueError:
+                        log.warning(f'Failed to parse region: {region}')
+                        continue
+
```
chris-simpson, line 568: So, on multi-extension ADs, this looks like it is fixing the same pixel regions on every extension, which is very unlikely to be what the user wants to do. A user is probably going to want to fix pixels in unmosaicked data, before interpolation happens, so do we need a more complicated syntax to specify an extension for each region?

chris-simpson (commented):
Apart from the specifics, a big question here is whether the user should be able to specify regions on *specific extensions*. Currently the same regions get fixed in all extensions and this is very unlikely to be what the user wants.

geminidr/core/parameters_preprocess.py
KathleenLabrie: No, that doesn't help.  It adds clutter.  Just 'eg. "450,521;430:437,513:533"'   But even then, I don't understand what this means.

450,521...  Is that a pixel *range*?   Since when a range gets expressed with a comma?   Or is that point?

Also, if the coordinates are 1-indexed Cartesian, with Fortran order, why does the "axis" parameter expects "0" as "first axis"?   First axis in Python sense, which then does not match the coordinates in "regions".    That doesn't look good to me.

geminidr/core/parameters_preprocess.py
KathleenLabrie: Overloaded input parameters are dangerous.  (re: using only regions for strings and file name)

chris-simpson: "Dangerous" how? I can't see any likely confusion between `my_regions.txt` and a string like `450,521;430:437,513:533`

KathleenLabrie: Messy in the code.

KathleenLabrie: At least in this specific case, the two inputs are strings.  That helps a bit, but it requires a fair bit of parsing and checking the nature of the string, "is this a file? or is this a region string?"

I'm the opposite of you, I don't like overloading input parameters.  I'm not against it, but I will always question it, to make sure that when we do it, we do it for the right reasons.

chris-simpson: I think it generally makes things easier for the user because they only have to remember the name of a single parameter and it's obvious to them whether it's a filename or a string (and it's not hard for the code to tell). Having parameters with similar names that do the same thing can lead to forgetting the names of parameters. If the code becomes messier, well that's secondary if it benefits the user experience. But having two parameters doesn't always make the code clean because sometimes you have to decide what to do if they're both set and should probably raise a warning (the code as written silently ignores the file, but I suggest that _both_ could be applied since they're not mutually exclusive and there might be advantages to this for the user).

KathleenLabrie: All I'm saying is think careful before just going for the knee-jerk reaction "I don't like input parameters, let's get rid of some."   

This is a case where I'm okay either way, but I push back to ensure it is a well thought decision, not just a reaction.

saimn: The main reason to have `regions` and `regions_file` is to raise a better error if you pass it a path to a non existing file. With only one parameter I would to parse the invalid path and the user would get a misleading error (or I would change the error message to a generic one).

> you have to decide what to do if they're both set and should probably raise a warning (the code as written silently ignores the file, but I suggest that both could be applied since they're not mutually exclusive and there might be advantages to this for the user).

Yes indeed. Depending on what we choose I will improve this.

geminidr/core/parameters_preprocess.py
saimn: > 450,521... Is that a pixel range? Since when a range gets expressed with a comma? Or is that point?

It's a single pixel. It's mentioned in the example, but I will also update the previous sentence to "colon-separated pixel
_coordinates_ or ranges".

> Also, if the coordinates are 1-indexed Cartesian, with Fortran order, why does the "axis" parameter expects "0" as "first axis"? First axis in Python sense, which then does not match the coordinates in "regions". That doesn't look good to me.

Right... well for regions parsing I reuse `cartesian_regions_to_slices` for which the choice was 1-indexed Fortran order. But for `axis` I just did the natural Python thing... As a Python user I would never think about specifying an axis in Fortran order, not sure why this choice was made for `cartesian_regions_to_slices` ?

geminidr/core/parameters_preprocess.py
saimn: > I'm not a fan of the long Config doc string, because it's inconsistent with all the other strings. There's a large discussion here about what showpars should do compared to the more extensive docstrings in the primitives (and how to access those docstrings from the shell).

Yes, I don't like either the long string here, but it was mostly to trigger the discussion. How users are expected to see the full docstring ? Maybe showpars could show it as well ?

geminidr/core/primitives_preprocess.py
```diff
@@ -533,6 +501,147 @@ def dilateObjectMask(self, adinputs=None, suffix=None, dilation=1, repeat=False)
...
+        timestamp_key = self.timestamp_keys[self.myself()]
+
+        if regions is not None:
+            regions = regions.split(';')
+        elif regions_file is not None:
+            with open(regions_file) as f:
+                regions = f.read().strip().splitlines()
+        else:
+            raise ValueError('regions must be specified either as a string '
+                             '(regions) or with a file (regions_file)')
```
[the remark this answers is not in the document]
saimn: Indeed, it could be useful to have the possibility to use both.

geminidr/core/primitives_preprocess.py
```diff
@@ -533,6 +501,147 @@ def dilateObjectMask(self, adinputs=None, suffix=None, dilation=1, repeat=False)
             gt.mark_history(ad, primname=self.myself(), keyword=timestamp_key)
         return adinputs
 
+    def fixPixels(self, adinputs=None, suffix=None, axis=None, regions=None,
+                  regions_file=None, use_local_median=False, debug=False):
+        """
```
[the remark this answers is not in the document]
saimn: Some of the other primitives in this file also have the explicit list (`darkCorrect`, `flatCorrect`, ...). I find this much more natural instead of having to unpack manually the params dict later. Btw the other primitives also use `suffix=None` which seems better than defining the real default value in multiple places (since it could change in the parameters module and get out of sync). Since the params are validated upstream the default value here does not matter, but at least it's good to have the explicit list in the signature instead of having the read the full code to find all the parameters a primitive can have (and docstrings are frequently incomplete).

geminidr/core/primitives_preprocess.py
```diff
@@ -533,6 +501,147 @@ def dilateObjectMask(self, adinputs=None, suffix=None, dilation=1, repeat=False)
...
+                        # If we have two axis with the same size, we should
+                        # use the deeper one. E.g. for images with a square
+                        # region, interpolation is done on lines.
+                        use_axis = np.where(
+                            region_shape == region_shape.min())[0][-1]
+                    else:
+                        if axis not in range(ext.data.ndim):
+                            raise ValueError('axis should specify a dimension '
+                                             f'between 0 and {ext.data.ndim}')
+                        use_axis = axis
```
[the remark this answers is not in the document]
saimn: Ok, there is a similar discussion above with @KathleenLabrie . So what you are suggesting would be the Fortran order also used by `cartesian_regions_to_slices` right ?

geminidr/core/primitives_preprocess.py
saimn: Yes, seems a good idea!

geminidr/core/primitives_preprocess.py
saimn: Good question. Another way to proceed would be to have a parameter to specify the extension to fix, for MEF ?

geminidr/core/primitives_preprocess.py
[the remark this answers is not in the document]
chris-simpson: Yes, `1` should be the first axis using the order in which the coordinates are supplied (i.e., `x`). We could allow/require a single character `x` or `y` (or `z`) as the parameter value to remove ambiguity, but maybe that's unnecessary.

geminidr/core/primitives_preprocess.py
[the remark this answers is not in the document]
chris-simpson: Specifying parameters in the call signature has been used sparingly when there aren't many and there's no chance of them being different for different instruments or changing. In many cases, `**params` is useful because many of the parameters are passed to a helper function (such as `fit_1D`) but it's just a style thing.

geminidr/core/primitives_preprocess.py
chris-simpson: I think the `regions_file` should be able to correct multiple extensions at once. For GMOS, for example, I wouldn't want to make the user create 12 files and then have to call the primitive 12 times with `-p extension=0 regions_file=fixpix0.txt`. But a parameter could indicate the extension for the `regions` parameter (if we're going to keep both, and apply both).

geminidr/core/primitives_preprocess.py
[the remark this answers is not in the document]
saimn: I went ahead and did this change.

saimn:
I addressed the various points, the only remaining one is to choose a syntax to specify the extension. With another separator ? 
`2@430:437,513:533` or `430:437,513:533@2` ? Other ideas ?

PR #109, opened 2020-12-30
Added codecov.yml with new threshold to prevent noisy red cov status

b1quint:
We noticed that recently the Coverage status was red when coverage dropped from 56.57% to 56.56%. This is a bit too much. I am adding the `codecov.yml` file with a tolerance of 1%. We might want to finetune this later if 1% seems to be way too permissive.

### DIFF
--- a/.github/codecov.yml
+++ b/.github/codecov.yml
@@ -1,0 +1,19 @@
+coverage:
+  status:
+    project:
+      default:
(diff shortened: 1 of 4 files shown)


b1quint:
Coverage went down by 0.8% and it is still in green state. Proved that it is working. Will put the tests back and merge.

PR #110, opened 2021-01-04
Remove padding from GMOSLongslit.normalizeFlat()

b1quint:
I found that the p.normalizeFlat() for GMOS LS is masking out 21 pixels at the edges of each amplifier of Hamamatsu data. I was wondering the following:
1) Shouldn’t these paddings be applied only to the edges of each detector instead? At least this is how it appears in the static BPMs.
2) I see in the docstring that there is a TODO asking if we should add this to the BPM. The new script/module to create BPMs for GMOS adds these padding so we could remove them from here if everyone agrees.

This PR simply removes the lines where the padding is added.

### DIFF
--- a/geminidr/gemini/lookups/timestamp_keywords.py
+++ b/geminidr/gemini/lookups/timestamp_keywords.py
@@ -9,6 +9,7 @@ timestamp_keys = {
     "addIllumMaskToDQ": "ADILLMSK",
     "addMDF": "ADDMDF",
     "addObjectMaskToDQ": "ADOBJMSK",
+    "addPaddingToDQ":  "ADPAD2DQ",
     "addReferenceCatalog": "ADDRECAT",
     "addVAR": "ADDVAR",
     "adjustWCSToReference": "CORRWCS",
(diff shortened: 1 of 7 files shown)


b1quint:
The `GMOSLongslit.normalizeFlat()` primitive should be applied to data that was not tiled, accordingly to @chris-simpson. Even if this is the case, I believe that this primitive should only do what it says. The next commits are a suggestion where I moved the padding on Hamamatsu detectors to a new primitive called `GMOS.addPaddingToDQ()`, which is called by `GMOS.addDQ()`.

chris-simpson:
I have two concerns about this.

First, this primitive masks the columns permanently and I'm not sure that's what we want. Those columns cause problems with `normalizeFlat` because they're very bright. But I think they're also very bright in imaging data yet aren't conspicuous in the final reduced image. If the reduction fixes them then we don't want them masked permanently. Maybe you've looked into this? But I don't recall seeing these spikes in reduced longslit data.

Second, the logic to decide which columns to mask needs improvement. You currently decide based on whether the data have been tiled or not. But tiled data might have been fully-tiled to a single extension (not 3) and untiled data might have 12, 6, or 3 extensions, depending on how many amps were used in GMOS. So you need to cater for these situations, probably by looking at the number of extensions, or the DETSEC. For the case where the detector has been tiled with `tile_all=True`, things are a bit more difficult although, since this is only for the Hamamatsus, we know the size of each detector so the locations are fixed.

b1quint:
Talk to Chris and it seems that this PR is not relevant for the two reasons mentioned above.

PR #111, opened 2021-01-11
Make imexam optional

saimn:
cc @jehturner


gempy/display/connection.py
```diff
@@ -86,7 +86,8 @@ def __new__(cls, viewer='ds9', path=None, use_existing=False,
                                    browser=browser)
 
         instance.logfile = 'imexam_log.txt'  # default logfile name
-        instance.log = imexam.util.set_logging()  # points to the package logger
+        if have_xpa:
```
jehturner, line 89: Why is the `imexam` logging only applicable for `ds9`? Is that just an `imexam` quirk?

setup.py
```diff
@@ -96,6 +94,7 @@
           'sqlalchemy',
       ],
       extras_require={
+          'all': ['ginga', 'imexam'],
```
jehturner, line 97: Hmm, what about `ds9`/`pyds9`? Ideally you would require *either* `ds9` or `ginga` for interactive use, but that's impossible to automate unless one of them is already installed. Also, it looks like `imexam` itself has a hard dependency on `ginga` (but not `pyds9`)... And also on `ipython` and `notebook`, in fact, which seems a bit much just to get `ds9` interaction :slightly_frowning_face:. The biggest issue was just the unconditional `imexam` import in order to use `geminidr`, but it would also be nice if `imexam` didn't pull all this stuff in, then it could just be a hard dependency (though it would still be one of only a few deps from Astroconda).

jehturner (commented):
I haven't tested your version, but if you're happy that it works as expected then fine.

The list of dependencies isn't complete though? What I have for run-time deps (including non-Python ones like sextractor and removing pytest for this purpose) is this:
```
    - asdf >=2.7
    - astropy >=4.2
    - astroquery >=0.4
    - bottleneck >=1.2
    - disco_stu >=1.3.7
    - future >=0.17
    - gemini_calmgr >=0.9
    - gwcs >=0.14
    - imexam >=0.8
    - jsonschema >=3.0
    - matplotlib >=3.1
    - numpy >=1.17
    - psutil >=5.6
    - pyerfa >=1.7
    - python-dateutil >=2.5.3
    - requests >=2.22
    - scipy >=1.3
    - scikit-image >=0.15
    - specutils >=1
    - sqlalchemy >=1.3
    - fitsverify >=4.17
    - sextractor >=2.8.6
```
I **think** `disco_stu` and `gemini_calmgr` are only required in order to follow the included documentation and are not actually dependencies of the code. Note that I haven't done a build & install test with these constraints yet.

jehturner:
I also have
```
    run_constrained:
    - ds9 >=8.0
```
and should probably add `pyds9` there too.

setup.py
saimn: The dependencies for imexam seems limited to numpy/scipy/astropy/matplotlib: https://github.com/spacetelescope/imexam/blob/d6f21d13d0e46a07de21512c0a25d5a97084a138/setup.cfg#L23-L51
I don't see ginga or ipython/notebook ?

gempy/display/connection.py
saimn: It's an imexam thing, but anyway our integration of ginga is done via imexam so that would be the same.

setup.py
jehturner: Hmm, that's just what `conda info` was telling me, but it looks like they've got rid of the `ipython` dependencies in the latest Python 3 builds, sorry (still has `ginga` though).

jehturner: https://github.com/astroconda/astroconda-contrib/blob/master/imexam/meta.yaml

jehturner: Sorry about the eventual S/N in that paragraph...

saimn: ginga seems optional (though missing in their setup.py) but conda does not have the concept of optional dependencies...

jehturner: Right, but/so usually if a dependency is optional, I would not include it in the conda package requirements. I'm not sure why they have done so here in that case.

jehturner: Maybe just to ensure that it has at least one back end available.

saimn: Yeah I don't know, it seems to come from an old version: https://github.com/astroconda/astroconda-contrib/pull/90

PR #114, opened 2021-01-18
Add -d / --doc option to show docstrings with showpars

saimn:
I also added some tests but for that I had to make `gempy.scripts` a package (adding ` gempy/scripts/__init__.py`), which you should be harmless.


KathleenLabrie:
This is brilliant.    Sync with master and merge.  Our users will really appreciate this new features.

PR #147, opened 2021-03-17
Refactor handling of calibration storage/retrieval in a more uniform manner

chris-simpson:


geminidr/f2/primitives_f2.py
```diff
@@ -25,8 +25,8 @@ class F2(Gemini, NearIR):
     tagset = {"GEMINI", "F2"}
 
     def __init__(self, adinputs, **kwargs):
-        super().__init__(adinputs, **kwargs)
         self.inst_lookups = 'geminidr.f2.lookups'
+        super().__init__(adinputs, **kwargs)
```
b1quint, line 29: Just curiosity, why did you move this line?

recipe_system/cal_service/tests/test_caldb.py
```diff
@@ -0,0 +1,94 @@
...
+
+        # Now remove it
+        caldb.remove_cal(cal_name)
+        assert len(list(caldb.list_files())) == 0
+
+        # Clean up
+        os.remove(caldb.name)
+
+
+@pytest.mark.skip("Need v1.0.0 of gemini_calmgr")
```
b1quint, line 73: An alternative could be [pytest.importorskip](https://docs.pytest.org/en/stable/reference.html#pytest-importorskip).

recipe_system/cal_service/remotedb.py
```diff
@@ -0,0 +1,191 @@
...
+        self.store_science = store_science
+        if not server.startswith("http"):  # allow https://
+            server = f"http://{server}"
+        self.server = server
+        self._calmgr = f"{self.server}/calmgr"
+        self._proccal_url = f"{self.server}/upload_processed_cal"
+        self._science_url = f"{self.server}/upload_file"
+        self._upload_cookie = upload_cookie or UPLOADCOOKIE
+
+    def _get_calibrations(self, adinputs, caltype=None, procmode=None,
```
b1quint, line 69: Does this have a particular reason to be private? Could we have some explanation in the docstring? (Same for the method below).

geminidr/f2/primitives_f2.py
saimn: I guess because the parent class' `__init__` will use it. 
Another way would be to make `inst_lookups` a class attribute.

b1quint:
[Review PR #147](https://trello.com/c/aQkJ8BDn/16-review-pr-147)

b1quint:
[Review PR #147](https://trello.com/c/aQkJ8BDn/16-review-pr-147)

geminidr/f2/primitives_f2.py
chris-simpson: The standard MDFs for each instrument are retrieved via the calibration service, and therefore they need to be added when the databases are set up. The databases are set up in `PrimitivesBASE.__init__()` but the locations of the MDFs are derived from `self.inst_lookups`. Therefore we need to define `self.inst_lookups` _before_ calling `super().__init__()`.

recipe_system/cal_service/remotedb.py
chris-simpson: The structure of the databases is that each one has a link to the next one in the list/cascade (whatever you want to call it). The private `_get_calibrations()` method only looks at its own database. The public `get_calibrations()` method (defined in the parent `CalDB` class) gets calibrations from its own database and then send any input files which didn't get a calibration on to the next database, so that's the one that should be used. It's the same for `_store_calibration()`.

PR #148, opened 2021-03-19
Closes #146 - Responsive size plots with window size 

### LINKED ISSUE #146: Responsive vertical/horizontal resizing for interactive traceApertures()
Right now, the interactive `traceApertures()` Web UI has a maximum width which is set via CSS. The plots' height also reaches a maximum and does not scale with the window's size, leaving lots of blank space. We want the plot's size to scale with the window to maximize its use.

b1quint:


b1quint:
I solved this issue partially with commits directly to `feature/interactive_refinements`. I could not find an easy and simple solution to add responsive behavior to the plot's height. So I am leaving this as it is, for now, to work on more urgent issues.

PR #174, opened 2021-03-31
Fix/159 region edition

### LINKED ISSUE #153: Setting maximum slide value raises FieldValidationError
The `RangeField` class accepts a `min` and a `max` parameter. When we run a primitive with `reduce -r primitiveCheese -p param=[max]`, this primitive fails with a FieldValidationError. For example: 

```
$ reduce -r traceApertures N20180508S0021_mosaic.fits -p max_shift=0.1 
   ... 
   ERROR - gempy.library.config.config.FieldValidationError: RangeField 'max_shift' (Maximum shift per pixel in line position
   ERROR - 	Valid Range = [0.001,0.1)) failed validation: 0.1 is outside of valid range [0.001,0.1)
   ERROR - 
   ERROR - Reduce instance aborted. 
```

We should reflect this behavior when creating Slides for Interactive Mode.

### LINKED ISSUE #159: "Regions" field do not work in interactive traceApertures
The interactive traceApertures() primitive contains a box below the plots that we can use to select regions of data used for fitting. This box is now completely ignored. It has no functionality.

### LINKED ISSUE #160: Key mapping does not work in interactive traceApertures()
The interactive traceApertures() primitive is supposed to accept Key Mapping commands. Right now, all the commands displayed in the mini-help panel on the bottom right are ignored.

b1quint:
Resolves #159, and resolves #160.

### DIFF
--- a/geminidr/interactive/controls.py
+++ b/geminidr/interactive/controls.py
@@ -264,9 +264,7 @@ class Controller(object):
                     self.unmask_handler()
 
             elif _key == 'p' or _key == 'P':
-                print('Pressing P')
                 if self.pointmask_handler:
-                    print('P has a handler!')
                     self.pointmask_handler(self.x, self.y)
 
             elif self.task:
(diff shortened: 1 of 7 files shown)


olyoberdorf (approved):
looks good, thanks for the extra cleanups

PR #219, opened 2021-04-10
add rejection of frames that cannot be sky subtracted to avoid contamination of the good ones.

KathleenLabrie:
Discussed at length with Chris.  Seems to work on my test sequence that highlight the problem in the first place.

### DIFF
--- a/doc/DRAGONS/changes.rst
+++ b/doc/DRAGONS/changes.rst
@@ -6,7 +6,46 @@
 Change Logs
 ***********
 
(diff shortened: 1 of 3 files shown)


chris-simpson:
The code looks OK but I don't like the term "sky table" in the log, as this has no meaning for the user. (And it's not really a table, it's just a list but we store it as a `Table` object because that's the only way to propagate it.) I'd prefer something like

Sky frames could not be associated to any input frames. Sky subtraction will not be possible.

and

{ad.filename} does not have any associated sky frames so cannot be sky-subtracted, moving to "no_skies" stream

[use an f-string like the cool kidz]

PR #221, opened 2021-04-13
refactored fit input data calculations into an inline method

olyoberdorf:
I created an inline method called _get_fit1d_input_data.

This takes in the basic inputs to calculate the fit input data and returns that ready for use in fit_1d.  For the interactive code, it can then build these inputs out into arrays to pass to the interactive fitter.  For the non-interactive version, these fit inputs can be used right away.

### DIFF
--- a/geminidr/core/primitives_spect.py
+++ b/geminidr/core/primitives_spect.py
@@ -343,6 +343,9 @@ class Spect(PrimitivesBASE):
                                 "1D spectrum")
                     continue
 
+            ### some sort of for loop with the warnings
+            # and build the indices to operate on
+
             def _get_fit1d_input_data(ext, exptime, spec_table):
                 spectrum = Spek1D(ext) / (exptime * u.s)
                 wave, zpt, zpt_err = [], [], []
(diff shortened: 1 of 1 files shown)


chris-simpson:
Er, you haven't actually replaced the code block in the non-interactive case with the new function.

Also, what's the purpose of that code under "# Hacking this out?"

There's also a lot of unnecessary propagation in the interactive case. You keep track of all the x and y units, but the visualizer simply prints that the x units are "nm" and the y units are "mag cm^2 electron / erg" instead of whatever the actual units are. Since the x and y labels are the same for all tabs, we should probably ensure the x and y units are the same for all extensions that produce a tab in the Visualizer. This is reasonable from a scientific basis since there will only be more than one tab for XD data, and the units really should be the same for all, so it makes sense to have some validation (which should apply for both interactive and non-interactive modes) and then you don't need a list of the units. This is a more significant rewrite since the validation encompasses much of what the first few lines in the `for ext in ad` loop (still duplicated) does.

olyoberdorf:
I'm not doing the non-interactive half until you're happy with the interactive side.  I'll make these updates and get back to you.

chris-simpson:
I'm not sure you can do one half in isolation for the reasons I outlined.

chris-simpson:
Having seen how this is developing, I put the creation of arrays in the "validation loop" for both interactive/non-interactive. I also changed the variable names from "pixels" and "masked_data" to "waves" and "zpt" which describe their actual nature in this particular primitive (the fitter can give them generic names like "masked_data" because it doesn't know or care what they are but I think the primitive should use specific and meaningful names). I haven't tested this (because I'm lazy) but I think it's done.

PR #232, opened 2021-04-14
adding provenance record for arcs - fixes 227

olyoberdorf:
Bias and flat handling has extra logic to record the calibrations in the provenance, beyond what the decorator does.

I added this for arcs.


chris-simpson:
Tests are failing. Some are regression tests, which is understandable since they will use the default inputs/references instead of the ones in the `release/3.0.x` branch, but some are not. Since all the tests pass on `release/3.0.x` I assume you've modified an outdated version of that branch. It would be helpful if you could merge in the more recent changes so we can confirm that no tests are failing unexpectedly.

PR #235, opened 2021-04-21
Removed old logic that was updating the list of editors

olyoberdorf:
The old logic was still in place to modify slider ranges and disable/enable for the set of editors per aperture.  We no longer have those and that was causing the stack trace error.


olyoberdorf:
Merging, failures are related to a different area of the code

PR #237, opened 2021-04-21
Bugfix/236 input box layout issue

olyoberdorf:
Input label text was overflowing for trace apertures.  This was mostly fixed by removing min/max text, but one of the parameters on the left was still too long.  I reworded it to a shorter message.


olyoberdorf:
Merging, test failures in unrelated code

PR #298, opened 2021-08-28
Move construction of science WCS based on arc into attachWavelengthSolution primitive

jehturner:
This factors out the scientific WCS construction from `distortionCorrect` and `QECorrect` into a single place and exposes the resulting WCS as part of each science exposure for use in other primitives and general reference/analysis and improved testing.

All of the existing tests have been updated accordingly and some have also been cleaned up a bit, eg. correcting unrelated problems with their `create_inputs_recipe`s. The `distortionCorrect` tests now test the correction on unmosaicked science data as well as mosaicked arcs and do so with a wavelength solution included in addition to the existing case that had distortion calibration only (ie. the canonical usage is now covered). A new red, GMOS-S Hamamatsu central-spectrum dataset has also been added in all these cases, to complement the blue, GMOS-N EEV full-frame one and spectroscopic BPMs are now applied, along with a few parameter tweaks for accurate results. Around 72 commented-out datasets for different spectral configurations have been eliminated, as they are mostly superfluous for this purpose and it was not possible to cross-reference results against them with new input files. A set of tests has been added, using the same data, for the new `attachWavelengthSolution` primitive, for both mosaicked & unmosaicked science data, checking the forward & inverse transforms. The output WCS from `distortionCorrect` is also now checked in both directions, not just the pixel values of the undistorted spectra.

The WCSs have been spot-checked fairly carefully and are accurate to something like hundredths of a pixel in the middle of the wavelength range and a fifth or quarter of a pixel at the ends of the ranges (from memory, but I do have some notes), plus or minus a minimal zero-point shift and some extra measurement uncertainty in the science exposures. A bug was fixed that caused a large shift in wavelengths where only distortion was calibrated and the approximate linear wavelength solution was used; the wavelengths in that scenario are now accurate within the ~1nm or so nonlinearity (plus any flexure), which is good enough to allow prominent lines to be identified confidently.

A bit of an [ugly hack](https://github.com/GeminiDRSoftware/DRAGONS/blob/e63f52e470ea6d69b463c9b350be0cacf9ea0458/geminidr/core/primitives_spect.py#L1066-L1085) remains in `distortionCorrect`, in order to preserve a possible ROI shift between arc & science exposures that gets prefixed as part of the distortion model; this could do with cleaning up later, perhaps by adding an ROI frame or something, TBD, which will be easier now there are more tests.


geminidr/core/primitives_spect.py
```diff
@@ -220,6 +221,312 @@ def adjustWCSToReference(self, adinputs=None, **params):
...
+
+                arc_detsec = arc.detector_section()[0]
+                ad_detsec = ad.detector_section()
+                if len_ad > 1:
+                    # We need to apply the mosaicking geometry, and add the
+                    # same distortion correction to each input extension.
+                    geotable = import_module('.geometry_conf', self.inst_lookups)
+                    transform.add_mosaic_wcs(ad, geotable)
+                    for ext in ad:
+                        # TODO: use insert_frame() method
```
jehturner, line 370: Fix this.

geminidr/core/primitives_spect.py
```diff
@@ -220,6 +221,312 @@ def adjustWCSToReference(self, adinputs=None, **params):
...
+                                        "same ROI. Continuing.")
+                            if 'sq' in self.mode:
+                                fail = True
+                            adoutputs.append(ad)
+                            continue
+                        # No mosaicking, so we can just do a shift
+                        m_shift = (models.Shift((ad_detsec.x1 - arc_detsec.x1) / xbin) &
+                                   models.Shift((ad_detsec.y1 - arc_detsec.y1) / ybin))
+                        m_distcorr = m_shift | m_distcorr
+                    # TODO: use insert_frame method
```
jehturner, line 459: Fix this.

geminidr/core/primitives_spect.py
```diff
@@ -220,6 +221,312 @@ def adjustWCSToReference(self, adinputs=None, **params):
...
+                    )
+                    output_frames.append(ext.wcs.output_frame.frames)
+
+                    # Shift science so its pixel coords match the arc's before
+                    # applying the distortion correction
+                    shifts = [c1 - c2 for c1, c2 in zip(ext.detector_section(),
+                                                        ext_arc.detector_section())]
+                    dist_model = (models.Shift(shifts[0] / xbin) &
+                                  models.Shift(shifts[1] / ybin)) | dist_model
+                    # TODO: use insert_frame method
```
jehturner, line 488: Fix.

chris-simpson:
This looks OK to me. However, I'll let you do the merge because the inputs and references will need to be copied from the branch directories into the standard `inputs/` and `refs/` directories.

jehturner:
Also you/we wanted to attach a wavelength solution when there's no distortion correction, didn't you? I would think that's not too fiddly but I'm not sure off hand; should be able to figure it out anyway and add another test. I suppose we should just copy whatever the arc has into the science exposure -- even if it's also just an approximate solution -- since that's what the user/recipe has specified by running the primitive?

chris-simpson:
Yes. Also `attachWavelengthSolution` is `core`-level so there may be instruments that don't require a distortion correction. Whether or not one is required should be handled by the caldb and the `procmode` of the arc so the primitive doesn't require any logic in terms of that. However, for now (since we only have GMOS) I suggest keeping a warning if there's no `distortion_corrected` frame in the arc.

jehturner:
We're going to merge this now, so that people can start on scientific testing, and I'll tidy up & change `attachWavelengthSolution` to use the wavelength solution from the arc when there's no distortion correction after doing a couple of other things.

PR #299, opened 2021-09-16
adding DRAGONS version to header of outputs

olyoberdorf:
In _write_final, if we are saving an output file, add a header keyword recording the version of DRAGONS used


KathleenLabrie:
This will be added only at the end of the recipe.  What about when I do writeOutputs() ?  What about intermediate products that some primitives create?

I don't actually have a solution.  We don't want to add this to astrodata since we plan to split it out.  Having recipe_system control it makes sense but I don't know when to add it.

Let's think ahead...

The versions that matter are the version of recipe_system (a bit) but mostly the version of geminidr.  Think of the use where a third-party package, say maroonxdr, is used.  We need that version number too.  It will use geminidr and maroonxdr, and "dragons".   geminidr can be DRGNSVER, I don't see splitting geminidr out of DRAGONS anytime soon.  Yet, I'd like to have a mechanism that would allow the external packages to add their version number too.
In the prepare primitive?   Some setVersion primitive?  I'd like to get Chris' opinion on this.

KathleenLabrie:
Another worry I have is what happens with this PR implementation when I do `reduce -r display` which does not change the data and therefore does not write an output.  I should not.  Will adding that header by default suddenly "change the file" and then force an output to be written?

olyoberdorf:
`writeOutputs` would work as well, but I really defer to you and Chris where this would best live.  In the `_write_final`, it is only saving the file if the filename changed, or a suitable suffix has been set.

PR #303, opened 2021-11-13
additional tests for find apertures with supernova data

olyoberdorf:
Got some extra test data from Monika and Chris.  May add more to this before merging, or update the primitive if we can find a problem input.


olyoberdorf:
closing for now to drop from jenkins - will reopen if/when we choose an algo

PR #304, opened 2021-11-17
Enh/gain readnoise propagation

chris-simpson:
This does the following: 
- sets the gain to 1 after ADUToElectrons
- warns if an image is being mosaicked/tiled with different gains on different extensions (using the tiling is being called by "display"), but sets the output gain to the mean of the input gains
- sets the output read noise when tiling/mosaicking to the mean of the input read noises
- warns if the gains are different when stacking images, but sets the output gain to the harmonic mean of the input gains
- sets the output read noise to the mean of the input read noises when stacking

In order to "set" the gain or read noise, there is a new decorator, "use_value_if_prepared", which will cause a descriptor to simply return the keyword value from the header (if it exists), rather than execute the code in the body of the descriptor. This is required because some instrument descriptors simply read a LUT and so the value returned is immutable.

### DIFF
--- a/geminidr/core/primitives_stack.py
+++ b/geminidr/core/primitives_stack.py
@@ -292,8 +292,8 @@ class Stack(PrimitivesBASE):
                 gains = [g[index] for g in gain_list]
                 # If all inputs have the same gain, the output will also have
                 # this gain, and since the header has been copied, we don't
-                # need to do anything!
-                if len(set(gains)) > 1:
+                # need to do anything! (Can't use set() if gains are lists.)
+                if not all(g == gains[0] for g in gains):
                     log.warning("Not all inputs have the same gain.")
                     try:
                         output_gain = num_img / np.sum([1/g for g in gains])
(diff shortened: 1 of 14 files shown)


geminidr/core/primitives_stack.py
```diff
@@ -153,25 +157,16 @@ def stackFrames(self, adinputs=None, **params):
...
+        gain_list = [ad.gain() for ad in adinputs]
+        rn_list = [ad.read_noise() for ad in adinputs]
 
         # Determine whether we can construct these averages
-        process_gain = all(g is not None and not isinstance(g, list)
-                           for gain in gains for g in gain)
-        process_rn = all(rn is not None and not isinstance(rn, list)
-                         for read_noise in read_noises for rn in read_noise)
+        process_gain = not any(g is None for g in flatten(gain_list))
+        process_rn = not any(rn is None for rn in flatten(rn_list))
```
olyoberdorf, line 165: just being pedantic, but why not
`if not None in flatten(rn_list):`

chris-simpson: Good idea. This code evolved and I missed the simplification.

PR #309, opened 2021-12-14
Combine darks with very different exp time

stinyanont:
The current version doesn't allow darks with e.g., 60.001 and 60.002 sec exposure times to be combined. This proposed change rounds the exposure time check to 2 decimal points to allow these insignificant exp time differences to be ignored.

### DIFF
--- a/geminidr/core/primitives_nearIR.py
+++ b/geminidr/core/primitives_nearIR.py
@@ -257,7 +257,7 @@ class NearIR(PrimitivesBASE):
         log = self.log
         log.debug(gt.log_message("primitive", self.myself(), "starting"))
 
-        if not all(dark.exposure_time() == adinputs[0].exposure_time()
+        if not all(round(dark.exposure_time(),2) == round(adinputs[0].exposure_time(),2)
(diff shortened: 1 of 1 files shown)


chris-simpson:
Thanks for your PR.

The small differences in exposure times for NIRI arises from differences in the number of coadds during the observation. An exposure time of 60.001s arises from a single coadd of 60s, whereas an exposure time of 60.002s arises from 2x30s coadds, so these frames are different. The NIRI Instrument Scientist recommends using darks with the same number of coadds as the observation from which the dark is going to be subtracted and therefore darks with different numbers of coadds (which manifests as a small difference in exposure time) should not be combined and DRAGONS is correct to prevent it. We should probably make this clearer with the exception that is being raised and this is something we have noted.

If you really want to combine these, then the `_varAdded.fits` files should have been written to disk as a result of the exception being raised, and you can simply run these through the `stackFrames` primitive (with `scale=False` and `zero=False`), which doesn't do the check.

stinyanont:
Makes sense. Thank you for taking a look.

Kaew Samaporn Tinyanont

PR #310, opened 2021-12-16
create AstroDataMixin to allow enhanced behavior of NDData-like classes

chris-simpson:
This fixes a MRO resolution problem with the `Spek1D` class from the `NDData` mixins. The base `NDAstroData` now has stuff related to memory-mapping, while `AstroDataMixin` contains convenience properties and methods that can be added in to `NDData`-like classes (such as `Spectrum1D`) to make them behave in the `AstroData` way we all know and love, as explained in the docstring.


olyoberdorf (approved):
Looks good, surprisingly straightforward

PR #311, opened 2021-12-17
function version of reduce and modified Reduce class to use it

olyoberdorf:
Pulled out the logic of the Reduce class into a reduce_data function:

* Reduce class uses the reduce_data function to implement runr() - which is what reduce_data was based on
* User arguments are passed to reduce_data as a dictionary as this is more natural than the list of tuples Reduce used
* reduce_data returns the list of output files


chris-simpson:
I just added some lines to `recipe_system.__init__()` so users can `from recipe_system import reduce_data` without the `reduction.coreReduce` cruft.

chris-simpson:
Suggestion from a user: could we log the parameters (and maybe user calibrations) when reduce gets going? This is something that can be done in the `reduce` environment, rather than in `PrimitivesBASE` since if a user is instantiating a `p` object directly they have to do all that stuff by hand. It will help with Helpdesk tickets in the future when users send in the log.

I think maybe this should be in `debug` mode rather than normal, but I'm not sure.

chris-simpson:
I meant use the existing log, with `log.debug()`, so it goes in the log. That's what the `--logmode` switch is for.

olyoberdorf:
oh - that makes more sense - changed and merged latest from master

PR #317, opened 2022-01-14
user parameter validation to check for unrecognized parameters (or pr…

olyoberdorf:
check for unrecognized parameters (or primitive:parameters)

This will fail-fast in Reduce() with a more readable error if the user:
* Passes a parameter name that is not present in any loaded Config definitions
* Passes a primitive:parameter where that parameter is not present in that primitive Config definition
* Passes a primitive:parameter where the primitive is not recognized (i.e. has no Config definition)
* Passes a thing with two or more colons that we can't interpret

### DIFF
--- a/geminidr/__init__.py
+++ b/geminidr/__init__.py
@@ -29,6 +29,7 @@ from astropy.io.fits.verify import VerifyWarning
 
 # new system imports - 10-06-2016 kra
 # NOTE: imports of these and other tables will be moving around ...
+from recipe_system.reduction.coreReduce import UnrecognizedParameterException
 from .gemini.lookups import keyword_comments
 from .gemini.lookups import timestamp_keywords
 from .gemini.lookups.source_detection import sextractor_dict
@@ -134,6 +135,8 @@ def cleanup(process):
     process.terminate()
 
 
+
+
 @parameter_override
 class PrimitivesBASE:
     """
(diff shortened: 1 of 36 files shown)


chris-simpson:
Some suggestions pushed. I abstracted some stuff, notably the search for "alternatives" (so later we can look for transpositions; I assume you intend something like this since you coded for the possibility that there might be more than one alternative).

I had to edit the test since the word "recognized" is used in all error messages whereas you had "found" sometimes and "recognized" sometimes.

I also explicitly used "parameter" for the parameter part of the user-supplied string rather than allowing "key" to drop through.

chris-simpson:
surely the `PrimitivesBASE._validated_user_parms is None` isn't needed in the logic you just changed, since this check requires that `self.user_params` has a value and so if `_validated_user_parms is None`, then it can't be equal to `self.user_params`. The tests were passing before this change so can you add a test that distinguishes between whether this is `==` or `!=`?

chris-simpson:
Urgh! You've had to pull the `find_similar_names()` function out of `_validate_user_params()` and monkeypatch it.

I assume this was to address my comment about the tests passing when the `!=` was (originally) `==`. But the existing test now _doesn't_ pass (I guess as a result of removing the `is None` check) if the `!=` is changed back to `==` so there seems to be no need for this new test and the unpleasantness you've needed to implement for it.

PR #323, opened 2022-01-26
Adding stackBiases primitive

DBerke:
This PR adds a new stackBiases primitive and two tests for it, and replaces stackFrames with stackBiases in gmos/recipes. The default rejection method is 'varclip', and the `zero` and `scale` option are not exposed to users.

It also tweaks stackDarks to remove `zero` and `scale` as user-facing options.

### DIFF
--- a/geminidr/core/parameters_ccd.py
+++ b/geminidr/core/parameters_ccd.py
@@ -2,12 +2,20 @@
 # in the primitives_ccd.py file, in alphabetical order.
 from gempy.library import config
 from astrodata import AstroData
-from . import parameters_generic
+from . import parameters_generic, parameters_stack
 
 class biasCorrectConfig(parameters_generic.calRequirementConfig):
     suffix = config.Field("Filename suffix", str, "_biasCorrected", optional=True)
     bias = config.ListField("Bias(es) to subtract", (AstroData, str), None,
                             optional=True, single=True)
+
+
+class stackBiasesConfig(parameters_stack.stackFramesConfig):
+
+    def setDefaults(self):
+        self.reject_method = 'varclip'
+        del self.zero
+        del self.scale
 
 
 class subtractOverscanConfig(config.core_1Dfitting_config):
--- a/geminidr/core/parameters_nearIR.py
+++ b/geminidr/core/parameters_nearIR.py
@@ -35,4 +35,7 @@ class separateFlatsDarksConfig(config.Config):
     pass
 
 class stackDarksConfig(parameters_stack.core_stacking_config):
-    pass
+
+    def setDefaults(self):
+        del self.zero
+        del self.scale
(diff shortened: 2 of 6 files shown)


DBerke:
I changed the test to use a set of biases of values (0, 1, 2, 2, 3) and to check that each gives the expected value, however: they all give the exact same expected value (1.6 for this set of bias values). Unfortunately I couldn't trace the algorithm far enough (it goes into a Cython function) to be able to tell if that was correct or not. Not sure if it'd be useful to make biases with a spread of values rather than a single one to see if that gives different results for the different rejection method.

chris-simpson:
Right, so varclip and sigclip will give the same result because varclip transforms to sigclip if there's no variance plane (which there isn't in your fake data). minmax also gives the same answer because the default values for nlow and nhigh are 0, which means nothing gets rejected (this seems silly since I believe they get ignored *unless* the rejection method is minmax) so if you pass nlow=1 and nhigh=1 to ``stackBiases`` then you should get a different answer for that parametriztion.

DBerke:
Ok, I'll try that with minmax. I could add a variance plane to the fake biases to ensure varclip gets run, even if it gives the same answer.

DBerke:
Ok, by adding a variance plane to the fake biases and telling 'minmax' to reject the highest and lowest pixels all three methods are now giving different results. 

I think 'minmax' is giving the expected value of 1.666... (5/3) from the inputs of [0, 1, 2, 2, 3], but I'll admit I don't understand how 'sigclip' and 'varclip' work well enough to tell if they're giving the correct values, even after following them as best I could through the code. But at 1.6 and 2 respectively they seem similar enough to 'minmax' that I'm guessing they're correct.

chris-simpson:
minmax with nlo=1, nhi=1 should indeed return 5/3.

sigclip rejects a pixel of value x if abs(x - m) > 3 * std (for 3-sigma clipping), where m is the mean (it's actually the median on the first pass, to avoid being overly affected by outliers) and std is the standard deviation, which is 1.0 for your sample, so no points get rejected and you get the actual mean of 8/5 = 1.6

varclip rejects a pixel of value x if abs(x - m) > 3 * sqrt(v) where v is the variance associated with the pixel, which is just x in the way you've made the var plane. So you should see that only the value 0 will get rejected and hence the result should be 8/4=2, which it is.

DBerke:
One thing I just thought of, would it be worth marking these as unit tests for Jenkins?

chris-simpson:
Go ahead, you can push that straight into `master`.

DBerke:
Ah, apparently "unit" is not among the Pytest markers (including our custom ones). It doesn't really fit any of the available ones either (it's not 'slow' or 'gmosls' or...), so I think we're fine with what we have now.

chris-simpson:
Looking at tox.ini, "unit" means anything that isn't marked with the other things: "integration_test", "gmosls", "regression", and "slow". So we're good.

PR #342, opened 2022-03-31
locking down relevant reduce parameters to be single-value

olyoberdorf:
locking down relevant reduce parameters to be single-value, producing appropriate error messages for extra unrecognized parameters rather than tacking them onto one of these options and then ignoring it


olyoberdorf:
per Chris, the change is fine but moving this into the reduce as function change and it will PR from there

PR #375, opened 2022-08-02
add minimal dither to makeFringeFrame

KathleenLabrie:
Seems to run as expected on the problematic data set, ie, it refuses to create a fringe on data not dithered by more than the minimal distance (currently set to 2 arcsec, might need to revisit to make larger)


KathleenLabrie:
The tests checked out.  Only the coverage was red, but I'll let that pass on this oldish release/3.0.x branch.  I'll merge it in.

PR #414, opened 2023-01-24
allow for optional N&S darks for Hamamatsu CCDs

KathleenLabrie:
tested it with and without N&S processed darks.  Seems to work as intended.


KathleenLabrie:
All checks passed, including coverage thanks to the new test.   Merging.

PR #415, opened 2023-01-26
Update `master` tests/docs to use conda-forge instead of Astroconda

jehturner:
This does what it says in the title. ~This is still pending a proper test run~, but I'm creating a draft PR to summarize the changes since it has to be done by Tues.

### DIFF
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -32,8 +32,7 @@ jobs:
       - uses: conda-incubator/setup-miniconda@v2
         with:
           auto-update-conda: true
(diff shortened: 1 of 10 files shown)


KathleenLabrie:
Speaking for myself, now is a good time to make a full switch to conda-forge, both for master and release/3.1.x.

jehturner:
I'll also merge the `conda_forge_3.1.x` branch into `release/3.1.x`, which is a very similar set of changes (I've redone 3 out of the 4 as straight cherry picks). And also `conda_forge_3.0.x` -> `release/3.0.x` when I'm finished with it.

Issue #426, opened 2023-03-07
caldb config claims that remote DB are configured when they are not


KathleenLabrie:
Branch: `release/3.1.x`

No remove DB is activated in `dragonsrc`.  Yet, when I do `caldb config`, I get this last line claiming that remote dbs are configured.  

```
(susd3.9_20230103) klabrie-ml2:playground klabrie$ caldb config

Using configuration files: ('~/.dragons/dragonsrc',)

/Users/klabrie/data/tutorials/gmosls_tutorial/cal_manager.db
  Type:  LocalDB
  Get:   True
  Store: True


Database file: /Users/klabrie/data/tutorials/gmosls_tutorial/cal_manager.db

There are remote dbs configured, meaning that remote calibrations will be downloaded.
```

This really needs to be fixed before we release 3.1.  It will cause massive confusion if released like that.

PR #428, opened 2023-03-07
Fixed accidental removal of caldb local/remote status check and a fix

olyoberdorf:
I'm not setup for a remote caldb.  Tested with local only and pretty confident this will work.

### DIFF
--- a/recipe_system/scripts/caldb.py
+++ b/recipe_system/scripts/caldb.py
@@ -213,16 +213,6 @@ class Dispatcher:
             print("   NB: The database does not exist. Please initialize it.")
             print("       (Read the help message about 'init' command)")
         print()
-        # local = True
-        # localdb = self.db
-        # seendbs = list()
-        # while localdb is not None:
-        #     if isinstance(self.db, RemoteDB):
-        #         local = False
-        #         localdb = None
-        #     else:
-        #         seendbs.append(localdb)
-        #         localdb = self.db.nextdb if self.db.nextdb not in seendbs else None
         if local:
             print(isactive)
         else:
--- a/recipe_system/scripts/caldb.py
+++ b/recipe_system/scripts/caldb.py
@@ -201,7 +201,7 @@ class Dispatcher:
         print(f"Using configuration files: \033[1m{filenames}\033[0m")
         print('')
         for db in dbs:
-            if db[0] != LocalDB.__class__:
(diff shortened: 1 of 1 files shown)


chris-simpson:
Hi Oliver, I hope you're well. I need some background here. You seem to be undoing a previous commit where you commented out a bunch of lines. What's the problem that these commits are fixing?

olyoberdorf:
Hi Chris, sorry for the drive-by PR.  This was something I looked at for Kathleen's issue #426, caldb config warns a remote DB is configured even though none are.

The original logic was detecting if we had any non-local caldbs and then printing a warning to the user.  This was a feature Kathleen wanted at some point.  But then I had commented out the logic, yet left in the warning.  This caused it to always warn a remote DB was in use even if it wasn't.

I'm not sure why I had removed the logic but as I commented it out (vs deleted) I believe it was a "quick fix" that I meant to revert with a proper fix and never did.

An alternative would be to keep the db walk commented out and also remove the warning about using a remote DB.  I'm also happy to do that, I have the code checked out and it's trivial and would be picked up by the PR automatically.

chris-simpson:
I'm still a bit confused because before this block of code is another block of code (starting `for db in dbs:`) which goes through the databases in the config file and sets `local`, which is what you say the (currently) commented block is also doing. This block (the `for` loop) walks through the databases by re-opening the config file, but the config file must have been opened by this script already and all the relevant information stuffed into `self.db` so it looks like these two blocks should be merged into one that does a walk through and both (a) determines whether there's a remote DB; and (b) prints stuff out regarding the name and the get/store status?

I think there also needs to be an `expand_filenames()` around the `if not exists(self.db.name)` since `db.name` retains the "~".

olyoberdorf:
Ah sorry, that sounds familiar now.  I fixed the bug in the logic comparing to `LocalDB.__class__` instead of to the `LocalDB` class itself.  I've also added and tested an `expanduser` for the existence check and I removed the dead code.

chris-simpson:
Now I understand better, I'm going to suggest [this](https://github.com/GeminiDRSoftware/DRAGONS/commit/0147647b275f1df62b7214552256fe3c9cd7954e)

It's a couple of lines longer but it doesn't have to call `parse_databases()` again because all the information is already in the `self.db` object.

Issue #365, opened 2022-06-01
The shifts for shiftImages should allow any number/type of spaces


KathleenLabrie:
Currently, the x, y shifts must be separated by exactly one space.  One space, two spaces, 10 spaces, a tab, whatever, should work just the same.  The split currently splits on " ", just one space.  A "re" could be use instead.

chris-simpson:
Fixed in this commit for spaces and tabs: 50d0ab82e11704f07e5d3a83df44b68ce3b3264e

PR #432, opened 2023-04-04
using re.sub to fix shifts for shiftImages should allow any number/ty…

phanicode:
\s+ greedily matches whitespace characters [\r\n\t\f\v ] from one and unlimited times


phanicode:
Issue #365

PR #439, opened 2023-06-29
Avoid conda race conditions within a Jenkins test run

jehturner:
Our Jenkins pipeline for testing DRAGONS seems to have been constructed with no regard for collisions between conda commands in parallel stages. In the past it has mostly worked anyway, but we have seen an increasing number of obscure failures in recent months, some of which do look very much like the result of multiple tests trying to update the Anaconda base installation at the same time, or trying to download the same package update at the same time. It is unclear why these problems have started occurring more often [update: see 4 Jul], but it could be due to changes in the behaviour of conda, or timing issues, including network glitches (which are another source of problems; see ITOPS-8891 on JIRA).

This PR reorganizes the pipeline in `master` a bit, so that the Anaconda/Miniforge base is updated by a single `Pre-install` stage and any package updates get downloaded there before the parallel test stages are executed. The package cache also gets cleaned out at the start (as in my `conda_err` branch), because we have seen numerous cases of it getting corrupted since April, which otherwise breaks subsequent test runs, until someone intervenes manually. This precuation could be a mixed blessing, since it could lead to more package downloads that are potential points of failure, but at least those failures should be recoverable, whereas a corrupt package cache currently leaves Jenkins defunct.

Unfortunately the new stage currently adds about 15 minutes to the execution time, mostly due to an extra `conda create` of the test env (rather than `conda update` in the base), but someone can look into optimizing that later -- currently I'm just trying to make Jenkins stable again. I think this overhead is larger than normal at the moment, because we've had to disallow the latest `astropy=5.3` due to an `NDData` bug, which makes the solver slower.

Note that creation of a test env in `Pre-install` (the slow bit) might not be essential, because I think the corruption of downloaded packages is temporary, ie. at some point the last conda invocation finishes overwriting the tarball and it is then valid again, as long as none of the test stages have tried to use it in the meantime. Consistent with this, I have seen some tests pass after an error about the astropy package being corrupt. Better safe than sorry though...

I have squashed most of the commits, to make this easier to merge into other branches, but have left the one that pre-creates the test env separate (see last para.).

This does not guarantee that simultaneous test runs on different branches won't collide, but that's less likely timing-wise and being able to run a single set reliably is a far bigger issue.

Having looked at the behaviour & overheads a bit, ~I now think we should consider using a completely separate base installation for each test stage instead~, since 1) moving from Anaconda to Miniforge has made it feasible, space- and speed-wise, 2) it turns out that the base installation overhead isn't that big compared with setting up the test env itself (especially if we have to clear out the package cache anyway) and 3) that should solve any race conditions between simultaneous test runs on different branches. Maybe I should look into that, which was less obvious when I started, but in the meantime this version should already be an improvement and the PR gives us somewhere to discuss the issue.

@KathleenLabrie, @DBerke

### DIFF
--- a/.jenkins/scripts/download_and_install_anaconda.sh
+++ b/.jenkins/scripts/download_and_install_anaconda.sh
@@ -32,6 +32,7 @@ fi
 
 echo ${PATH}
 which conda
+conda clean -i -t -y  # in case of corrupt package cache from previous run
 conda update --quiet conda
 # These 2 channels need removing if testing old branches has reinstated them:
 conda config --remove channels http://ssb.stsci.edu/astroconda || :
--- a/.jenkins/scripts/setup_agent.sh
+++ b/.jenkins/scripts/setup_agent.sh
@@ -3,17 +3,8 @@
 
 set -eux
 
-# Cleanup
+pwd
 git clean -fxd
-mkdir plots reports
-if [[ -n "${DRAGONS_TEST_OUTPUTS-}" ]]; then
-    if [[ -d "${DRAGONS_TEST_OUTPUTS}" ]]; then
-        echo "Cleaning previous test results in ${DRAGONS_TEST_OUTPUTS}"
-        rm -r ${DRAGONS_TEST_OUTPUTS}
-    else
-        echo "Skip delete unexisting ${DRAGONS_TEST_OUTPUTS}"
-    fi
-fi
 
 source .jenkins/scripts/download_and_install_anaconda.sh
 
--- a/.jenkins/scripts/setup_dirs.sh
+++ b/.jenkins/scripts/setup_dirs.sh
@@ -1,0 +1,20 @@
+#!/bin/bash
+# -*- coding: utf-8 -*-
+
+set -eux
+
+# Cleanup
+pwd
+git clean -fxd
+mkdir plots reports
+if [[ -n "${DRAGONS_TEST_OUT-}" ]]; then
+    if [[ -d "${DRAGONS_TEST_OUT}" ]]; then
+        echo "Cleaning previous test results in ${DRAGONS_TEST_OUT}"
+        rm -r ${DRAGONS_TEST_OUT}
+    else
+        echo "Skip deletion of inexistent ${DRAGONS_TEST_OUT}"
+    fi
+else
+    echo "DRAGONS_TEST_OUT is not set, so not deleting it"
+fi
+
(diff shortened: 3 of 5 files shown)


jehturner:
PS. I have also added a `conda` label to each node in Jenkins (except the on-line one in Hawaii, since we currently don't want to start triggering jobs there). This  signifies that conda needs installing on the machine and seemed like the best way to avoid potential maintenance headaches associated with using existing labels.

KathleenLabrie:
That's a nice summary of the issue and the solution.  Feel free to merge (you can merge yourself, right?) when you are ready.  It will be needed in `master` and in `release/3.1.x`.

DBerke:
I can't say I understand every detail, but I get the gist of what's been changed, and it sounds like this should get Jenkins back on track. An extra 15 minutes per test run isn't that much in the grand scheme of things, and we can start to think about optimization when we have it working again and some room to breathe.

jehturner:
A bit more on this ... I think the main change in `conda` that was causing our tests to break more often was a reorganization of how channel indexing works (eg. https://github.com/conda/conda-index/blob/a6e4002dd8eac1e0de8c06aa76e07c53a83135e7/README.md#summary-of-changes-from-the-previous-conda-build-index-version & other refs). It is now completely intolerant of any concurrency; all it takes is 2 simultaneous `conda create` commands for _different_ environments (which try to update common cache entries simultaneously) to corrupt the index until someone runs `conda clean`. Fortunately, after some fiddling, I found that a new experimental locking option does seem to resolve the problem. There is also a `--use-index-cache` option, to avoid cache updates, which could have worked if there were a way to tell `tox-conda` to pass it only after the first invocation in `Pre-install`, but the cache has to be updated the first time, otherwise some of the packages cannot be found.

The restructuring in this PR _also_ addresses some other failure modes, notably due to multiple `conda create` commands for different envs downloading the same package updates at the same time, corrupting their tarballs (at least temporarily), and multiple tests trying to run `conda update` on the same base env at the same time. Those kinds of things have been a lower-level nuisance in the past.

Regarding my last (late-night) paragraph in the initial description above, I realized that the problem with independent installations would be how to avoid downloading all the necessary scientific packages at every test stage (x11), which would not only be inefficient but also fragile, exacerbating the problem we sometimes see with network time-outs. In considering how to deal with that, I ended up circling back to something not too different from the scheme I have here, but more convoluted, so I am leaving optimization as a later exercise. My latest thought on the matter is that it's not really necessary for each test stage to install the same set of packages independently in its own test environment (especially when they are not _really_ isolated) -- we could install a common test env once, at the start -- but that's just how `tox` works. A glance at `astropy`'s CI suggests that it does the installation separately and then tells `tox` to omit that step when running tests. Anyway, this all needs redesigning properly by someone with more time than I have right now. In the meantime I'll merge these changes so it actually works again :slightly_smiling_face:.

PR #451, opened 2024-04-23
Add marker and palette color for updated mask list.

teald:
The palette in Fit1D was expanded with a new mask, but the change did not propagate forward the additions to the markers and palette (colors) list in `InteractiveModel`. This fix adds the `"square"` marker and `"orange"` color for the `"threshold"` mask.


KathleenLabrie (commented):
I tested the tools. The warning is gone.

In the future, can we keep the non-critical/preference style changes to a minimum and maybe in their own commit to help make clear which changes actually matter.

PR #452, opened 2024-04-23
Set wavecal wavelength labels relative to spectrum

teald:
Previously, the wavelength calibration set the label position based on, depending on the circumstances, the plot's y-axis or the spectrum's heights. This takes the peak heights and applies a padding to place the text.

This hopefully solve a problem with the bokeh server stalling when wavelengths are added to the plot by a user or the calibrator.

### DIFF
--- a/geminidr/interactive/fit/wavecal.py
+++ b/geminidr/interactive/fit/wavecal.py
@@ -552,6 +552,25 @@ class WavelengthSolutionPanel(Fit1DPanel):
 
         return new_model
 
+    @staticmethod
+    def _get_plot_ranges(plot: figure):
+        """Get the start and end of the x and y ranges of a plot."""
+        x_range = plot.x_range
+        y_range = plot.y_range
+
+        ranges = {
+            "x": (x_range.start, x_range.end),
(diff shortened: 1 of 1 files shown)


chris-simpson:
OK, this sorts out the CPU usage but causes problems with the plots, as the attached screenshot shows when I zoom in to regions of the two plots (the labels in the top plot are mostly lost off the top).

If I calculate `height` based on refplot line intensities or spectrum data within the plot's `x_range` things are improved but labels can still be truncated off the top of the plot. I can't think how to fix this right now.

Screenshot 2024-04-26 at 3 16 23 PM

teald:
I'm a bit more worried about the CPU issue affecting people trying to use the interactive mode. I've done some experimenting this morning and with the current setup fixing this label height issue is nontrivial. Ideally, it would only overflow if the space between the peak/line is not enough to fit the text, but it's not straightforward to fetch the correct label dimensions in every place they're needed.

I think it may be best to integrate these changes sooner rather than later, alongside the other interactive fixes. Hopefully I can figure something out sooner rather than later 🤞

PR #453, opened 2024-04-24
findApertures input fixes

teald:
# Slider ranges can now be explicitly set

`config.RangeField` had no means of communicating a desired slider range when constructing a slider. Frequently, the slider should only span the min/max range, but `PrimitiveVisualizer.build_text_slider` would default to `end=10` if the maximum value or twice the value was less than 10. Precisely, for situations where the initial value is provided ([interactive.py line 1082](https://github.com/GeminiDRSoftware/DRAGONS/blob/1f65fbf60a7e780dd540298f0d2370f388907e2a/geminidr/interactive/interactive.py#L1082)):

```python
# if min/max value is None/Falsey, use a default.
start = min(value, min_value or 0)
end = max(value, max_value or 2 * value, 10)
```

Since this default is currently used throughout the interactive module, I've left it in for now. It will need to be changed when the interactive refactor happens.

Instead, both `build_text_slider` and `RangeField` now accept two new keyword-only arguments, `fix_start_to_min` and `fix_end_to_max` (both named to communicate the slider functionality). They are `False` by default, preserving the original behavior wherever necessary.

I chose to make them positional-only to avoid adding to positional argument function calls, since both `RangeField` and `build_text_slider` are already crowded with positional and [normal] keyword args.

## Fixes to slider stepping

The step for the slider, when using a float, was set to `0.1`. For some sliders this is greater than or equal to the total range, so now it will be `min(0.1, (end - start) / 100)` instead. This is enough for relatively smooth slider movement regardless of the start/end values.

## Other notes

This has been applied to sliders in `traceApertures` as well.

# Adds `reinit_live` fix to nested `TextHandler` class

In `PrimitiveVisualizer`, there is a `TextHandler` class (unfortunately, defined in the function) that was causing problems by trying to access the `PrimitiveVisualizer.reinit_live` without a reference to the `PrimitiveVisualizer` instance in-scope. Instead of a clever solution, like passing the instance, I'm just passing the value of `reinit_live`, since this `TextHandler` class will be refactored later.

### DIFF
--- a/geminidr/core/parameters_spect.py
+++ b/geminidr/core/parameters_spect.py
@@ -245,7 +245,7 @@ class findAperturesConfig(config.Config):
     use_snr = config.Field("Use signal-to-noise ratio rather than data in "
                            "collapsed profile?", bool, True)
     threshold = config.RangeField("Threshold for automatic width determination",
-                                  float, 0.1, min=0, max=1)
+                                  float, 0.1, min=0, max=1, fix_end_to_max=True)
     interactive = config.Field("Use interactive interface", bool, False)
     max_separation = config.RangeField("Maximum separation from target location (arcsec)",
                                        int, None, min=1, inclusiveMax=True, optional=True)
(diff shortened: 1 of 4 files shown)


KathleenLabrie:
The slider works well.  It's much cleaner this way.  thanks.

PR #463, opened 2024-06-07
Extend `logutils.config` to accept additional handlers.

### LINKED ISSUE #462: Extend `logutils.config` to Support Custom Handler Integration
**Issue Description:**

Currently, I'm able to hack adding custom log handlers to capture logging output from DRAGONS except for messages generated within the primitives section. This limitation means that significant log data essential for debugging and monitoring are not captured, especially when these logs are required on external systems like the GOATS frontend.

**Proposed Solution:**

I propose extending the `logutils.config` function to accept an additional argument named `additional_handlers`. This argument would be designed to accept either a single logging handler or a list of handlers. This will allow GOATS or other projects to use their own handlers if needed and capture all the output.

davner:
Closes #462.

This pull request extends the `logutils.config` function by enabling the attachment of custom logging handlers. This update is essential for the GOATS project, as it allows for all DRAGONS log messages to be sent and displayed in real-time on the frontend via a WebSocket connection.

### DIFF
--- a/gempy/utils/logutils.py
+++ b/gempy/utils/logutils.py
@@ -121,7 +121,8 @@ def get_logger(name=None):
         customize_log(log)
     return log
 
-def config(mode='standard', file_name=None, file_lvl=15, stomp=False):
+def config(mode='standard', file_name=None, file_lvl=15, stomp=False,
+           additional_handlers=None):
     """
     Controls Dragons logging configuration.
 
@@ -136,8 +137,12 @@ def config(mode='standard', file_name=None, file_lvl=15, stomp=False):
     file_name : <atr>
           filename of the logger
 
-    stomp: <bool>
+    stomp : <bool>
           Controls append to logfiles found with same name
+
+    additional_handlers : `logging.Handler` or <list>
+        An initialized handler or a list of initialized handlers to be added to the
+        root logger.
 
     Returns
     -------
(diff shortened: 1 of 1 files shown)


gempy/utils/logutils.py
```diff
@@ -203,6 +208,14 @@ def config(mode='standard', file_name=None, file_lvl=15, stomp=False):
         log_handler.setLevel(file_lvl)
         rootlog.addHandler(log_handler)
 
+    # Attach additional handlers if provided.
+    if additional_handlers is not None:
+        if isinstance(additional_handlers, list):
```
teald, line 213: This will fail on some iterables. I think [`collections.abc.Iterable`](https://docs.python.org/3/library/collections.abc.html#collections.abc.Iterable) may be better to check again if keeping `isinstance`. It has some limitations noted in the linked docs. It recommends a more generic:
```python
try:
    iter(additional_handlers)  # Ensures additional_handlers is iterable
    is_iter = True
except TypeError:
    is_iter = False
  ```
  Or something similar, just being explicit with `is_iter` for clarity. Since it's doubtful any weirdly complex iterators would be passed I think it's fine to use `isinstance(..., collections.abs.Iterable)`, but I'll leave that up to you.

Comment (teald) [reply gempy/utils/logutils.py:213]: Alternatively:
```python
try:
    iter(additional_handlers)
except TypeError:
    additional_handlers = [additional_handlers]
finally:
    # ... do things with ensured-iterable additional_handlers ...
```

davner: Ooo good catch. I'll make the change.

teald (approved):

PR #475, opened 2025-03-05
merge Enh/fixlogging - restore API compatibility between gempy logger and python logger, tidy up

phirstgemini:
* Restores API compatibility of the `gempy.utils.logutils` logger with the python native logger, which had been broken by introduction of custom `log.methods()` that assumed the only argument passed would be a string, which is not the general case. These had been introduced to provide functionality to only put the log level name in the console log for WARNINGs and above and DEBUGs (ie not for the info type messages). This is now implemented with a custom formatter class in the console log handler.
* Adds tests for gempy logutils
* Tidies up and documents the logutils.py module somewhat
* Removes the log_traceback code from rs_utilities and replaces calls to it to use python native logger `excinfo=True` functionality instead.

* This PR currently does NOT change the custom log levels that are added to the logger.

### DIFF
--- a/gempy/utils/logutils.py
+++ b/gempy/utils/logutils.py
@@ -164,7 +164,6 @@ class DragonsConsoleFormatter(logging.Formatter):
         return ' ' * (self._indent_level * self._indent_width)
 
     def _init_formatters(self):
-        print("Init formatters")
         self._short_fmt_str = self._indent_str() + '%(message)s'
         self._long_fmt_str = self._indent_str() + '%(levelname)s - %(message)s'
         self._short_formatter = logging.Formatter(self._short_fmt_str)
@@ -176,12 +175,9 @@ class DragonsConsoleFormatter(logging.Formatter):
 
     def format(self, record):
         # Levels for which to emit short form
-        print(f"{record.levelname=} {self.short_levels=}")
         if record.levelname in self.short_levels:
-            print("short")
             return self._short_formatter.format(record)
         else:
-            print("long")
             return self._long_formatter.format(record)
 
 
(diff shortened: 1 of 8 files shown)


phirstgemini:
Not sure why it's complaining about CI. It shows green on jenkins.

gempy/utils/logutils.py
```diff
@@ -43,64 +70,75 @@ def add_filter(filt, log):
...
-    setattr(log, 'critical', ccritical)
-    setattr(log, 'error', cerror)
-    setattr(log, 'warning', cwarning)
-    setattr(log, 'status', cstatus)
-    setattr(log, 'stdinfo', cstdinfo)
-    setattr(log, 'info', cinfo)
-    setattr(log, 'fullinfo', cfullinfo)
-    setattr(log, 'debug', cdebug)
 
+    # If this logger has already been customized, this is a no-op
```
teald, line 90: (nitpick) If you wanted to preserve this behavior without touching the object, `Logger` objects are hashable so you can use `functools.cache` (or `functools.lru_cache`) so that you get the no-op behavior with just a decorator:
```python
from functools import cache

@cache
def customize_logger(...):
    ...
```
Adding an attr here gets the job done, no need to change it unless you'd like to. I just didn't know `Logger`s were hashable until I tried it just now and wanted to mention it. It looks like it's just the default `__hash__` implementation ~based on the object id.

teald (approved):
Great updates! I especially like that `exc_info` is no longer handled by its own function and that log utils had some tests now. I had just one non-critical comment. Thanks for the work, lgtm.

gempy/utils/logutils.py
phirstgemini: Thanks, that didn't occur to me. My instinct is to leave it as it is, mostly so that if the next person to touch the code (possibly me) doesn't remember the cache stuff, it's simple and explicit as is. It does also allow for user code to check if it's has a "stock" logger or a "dragons customized" logger very clearly.

It did occur to me that it could be argued that there should not be a get_logger() function here - the rationale being that the script using it (ie the thing where __name__ = '__main__' should in principle do:
```log = logging.getLogger()
logutils.customize_logger()```
and from them on, code should just do log = logging.getLogger() and they will get the singleton customized logger. However, having it like it is is maybe more what people not so familiar with it expect and this seemed a good compromise of improving things without changing too much at once...

PR #476, opened 2025-03-06
Ensure clean subprocess termination and resource cleanup (GOATS-659)

davner:
### Summary of Changes:
- Implemented an exit strategy in `loop_process` by handling `None` commands to break.
- Modified `ETISubprocess.terminate()` to gracefully stop the process and close queues.
- Reset the singleton ETISubprocess instance on termination.
- Prevented zombie processes and orphaned multiprocessing forks in long-running applications like GOATS when shutting down.

Hello DRAGONS team!

We’ve observed lingering multiprocessing forks and resource trackers when running DRAGONS reductions in GOATS. These changes address the issue for long-running applications and improve resource management.

### Details:
These changes address an issue where DRAGONS’ `runr` process leaves behind zombie and orphaned multiprocessing forks when used in long-running applications like GOATS or Jupyter notebooks. The root cause is that DRAGONS relies on `atexit` for cleanup, but in environments where the interpreter is not explicitly exited (such as persistent Django applications or even Jupyter notebooks), `atexit` is never triggered. This results in lingering child processes that accumulate over time.

### `loop_process` Changes:
By ensuring `loop_process` can be cleanly exited, GOATS can properly terminate workers without having to force terminate.

### `ETISubprocess.terminate` Changes:
- Puts in `inQueue` `None` which cleanly breaks the `while True` loop.
- The process can possibly cleanly exit without force termination.
- The queues are now closed to free up system resources.
- The singleton instance is set to `None` to reset stale references.

### Note
If changes to the cleanup process, including resource deallocation and singleton resetting, are not desirable, please consider at minimum adding the ability to cleanly break out of `loop_process`. As it stands, GOATS has to force terminate this.

### DIFF
--- a/gempy/eti_core/eti.py
+++ b/gempy/eti_core/eti.py
@@ -13,13 +13,20 @@ def loop_process(in_queue, out_queue):
     Parameters
     ----------
     in_queue : multiprocessing.Queue
-        (add docstring)
+        A queue from which commands are received for execution.
     out_queue : multiprocessing.Queue
-        (add docstring)
+        A queue to return the results of the executed commands.
+
+    Notes
+    -----
+    - If `None` is received, it breaks out of the loop and exits cleanly.
     """
     while True:
         try:
             cmd = in_queue.get()
+            # Offer way to cleanly break out of while loop.
+            if cmd is None:
+                break
             try:
                 result = check_output(cmd, stderr=STDOUT)
             except CalledProcessError as e:
@@ -44,8 +51,36 @@ class ETISubprocess:
                                    args=(self.inQueue, self.outQueue))
             self.process.start()
 
-        def terminate(self):
-            self.process.terminate()
+        def terminate(self, timeout=2.0):
+            """Terminate the subprocess gracefully (if possible), or force-kill if needed.
+
+            Parameters
+            ----------
+            timeout : float
+                Number of seconds to wait before forcing termination.
+            """
+            # Send quit message to loop.
+            self.inQueue.put(None)
+            self.process.join(timeout=timeout)
+
+            # If still alive, force-terminate.
+            if self.process.is_alive():
+                self.process.terminate()
+                self.process.join()
+
+            self.process.close()
+
+            # Close queues.
+            try:
+                self.inQueue.close()
+                self.inQueue.join_thread()
+                self.outQueue.close()
+                self.outQueue.join_thread()
+            except Exception:
+                pass
+
+            # Reset instance, ensures new subprocess is created.
+            self.instance = None
 
     instance = None
 


chris-simpson:
I understand that ETI subprocesses will linger if their parents aren't exited. That's kind of a necessity because the subprocess makes a duplicate of its parent and so must be created when the memory footprint is small. What you absolutely don't want is to kill the child process once the parent has become large in memory and then have to create a new child of that process.

I infer (perhaps wrongly) that you intend to explicitly kill the child ETI process once GOATS has executed a recipe because completing that reduction process doesn't trigger `atexit()`.

In GOATS, is the ETI process a child of the specific process executing that recipe, or is it a child of GOATS? If it's a child of that specific recipe execution process, then I don't understand why that process doesn't trigger `atexit()` when the recipe is completed (or why GOATS doesn't explicit cause the interpreter to exit). And if it's a child of GOATS then it _should_ linger so you don't have to create a new child later which may have a larger memory footprint.

I infer (again, perhaps wrongly) that your new version of the `terminate()` method is there because you feel that sending a `None` command to break out of the loop is cleaner than calling the `terminate()` method of the subprocess itself?

davner:
> I infer (perhaps wrongly) that you intend to explicitly kill the child ETI process once GOATS has executed a recipe because completing that reduction process doesn't trigger atexit().

I understand why you would think that. GOATS spawns worker threads that do not shutdown until GOATS is terminated. At this time, I am killing the child ETI process at shutdown, not at completion of a recipe. When GOATS is being shutdown, a normal exit is not possible for the worker threads. So what I have done is written middleware that gets the `ETISubprocess` singleton, passes the sentinel value to break out of `loop_process` and that calls the `ETISubprocess.terminate()` when a worker is shutting down, which happens when GOATS is called to shutdown.

> I infer (again, perhaps wrongly) that your new version of the terminate() method is there because you feel that sending a None command to break out of the loop is cleaner than calling the terminate() method of the subprocess itself?

Correct, I do believe it is cleaner. Does it really make a difference...not really. I don't like forever while loops without a way to terminate it. This provides an easy shutdown mechanism. In my middleware, I try to do a "clean" shutdown, however, without a way to interrupt the loop, I instead call `terminate()` method of the subprocess itself. 

I added resource cleanup steps to close queues and processes. I believe it’s best to be explicit about cleaning up after ourselves, especially in long-running worker environments, like GOATS, so we don’t leave any stray processes or open resources behind.

chris-simpson:
Thanks for clarifying. I'm not an expert on this and certainly not on the lower-level stuff but the changes seem desirable and the commit looks fine so I'll go ahead and merge. (I note the build failure but this doesn't look to be due to the commit.)

PR #479, opened 2025-03-08
Add DRAGONS software name and version number headers in standardizeHeaders so that they appear in reduced data files for GOA

phirstgemini:
Add `PROCSOFT` and `PROCSVER` headers to PHU in `geminidr/core/primitives_standardize.py::standardizeHeaders` to facilitate support of DRAGONS reduced data in FitsStorage and GOA.

I added these two keywords to the ignore list in `ad_compare` so that a) we don't have to regenerate the entire reference file set _now_, and b) we we don't have to regenerate the entire reference file set with every release. We could remove PROCSOFT from that exclusion once it's made it's way into all the files, if we want to. I added a separate test for `standardizeHeaders` to check these appear as desired. 

I also put a comment in `reduce.py` so that if / when we change how the DRAGONS version is defined, we remember to update the `standardizeHeaders` primitive to use the same as what `reduce.py` announces on startup.


phirstgemini:
This would make PR #299 obsolete.

phirstgemini:
Having heard no objections, I'm going to go ahead and merge this.

PR #481, opened 2025-03-11
register ETISubprocess.terminate() with atexit when it is created, so…

chris-simpson:
… there's only one call if multiple PrimitiveBASE objects are created

This fixes the issue made apparent by #476 that multiple calls to terminate the ETI Subprocess could be stored in the `atexit` reigstry.


davner:
I'm going to test this in GOATS today.

davner (approved):
Looks good! Tested with GOATS and the test scripts I wrote to check for zombie/orphans and none exist. Thanks for getting this fixed.

chris-simpson:
Thanks. I'm glad you were able to identify and diagnose this.

PR #482, opened 2025-03-11
update Reduce API docs 

phirstgemini:
* Note about customize_logger()
* Say that you need to use the if __name__ == __main__ idiom in any scripts calling Reduce(). Explanation cribbed from Chris in https://noirlab.atlassian.net/browse/EXTHD-437


phirstgemini:
@KathleenLabrie , I'd like to go ahead and merge this to get it off the list. Any objections?

KathleenLabrie (approved):

PR #483, opened 2025-03-11
Improved telluric correction, spectral resampling, and wavecal/ATRAN refactor; plus F2 wavecal

chris-simpson:
This is a biggie, combining my work over the past year or more. It features:

1. Further development of the primitives `fitTelluric()` and `telluricCorrect()`. These broadly work but need a little polishing.
2. Improvement to F2 wavelength calibration by creating a `peak_to_centroid()` function that estimates the pixel location of a line's centroid (the actual location of the reference wavelength) from the peak (found by the code). Since the peak varies with smoothing scale, the `fwidth` parameter is set by the slit width to ensure consistency.
3. Improved resampling in `Spect.resampleToCommonFrame()` that allows for additional resampling options (log-linear resampling is already coded but needs testing) and allows XD orders to be resampled to their own wavelength scales.
4. Fixes a bug in `Spect.resampleToCommonFrame()` that caused the output wavelength scale to be incorrect.
5. Causes `trim_spectral=True` in `Spect.resampleToCommonFrame()` to behave as intended and documented and return the intersection of the spectral coverages, not the coverage of the reference spectrum.
6. Refactored parts of the wavecal code so that a Chebyshev1D is always returned as the initial model, with a domain that describes the illuminated part of the detector. This allows some of the code in the ATRAN modelling to be removed.
7. Reference sky-line spectra are now constructed on the fly to the precise resolution determined by the `LineSpreadFunction` instance, instead of being selected from a list of pre-convolved spectra.

Reference files in the following directories will need to be moved into the appropriate `refs` directories
- geminidr/core/test_spect/refs_enh_telluric2
- geminidr/f2/longslit/test_determine_wavelength_solution/refs_enh_telluric2
- geminidr/gnirs/longslit/test_resample_2d/refs_enh_telluric2
- geminidr/gnirs/recipes/sq/test_xd_spect/refs_enh_telluric2
- geminidr/gnirs/recipes/sq/test_ls_spect/refs_enh_telluric2


chris-simpson:
Merging because nobody wants to review all this.

PR #484, opened 2025-03-12
fix double logging bug

phirstgemini:
I made a booboo in my last update in that I failed to appreciate a thing about list mutability.


phirstgemini:
I'm going to just go ahead and merge this, because it's a small bugfix change and it's all green, and life's too short...

PR #486, opened 2025-03-13
Fix a bug returning telluric models to the primitive. Plus force a parameter change in resampleToCommonFrame

chris-simpson:
Two things.

1. The `TelluricCalibrator` object was returning the wrong values as coefficients of the continuum model fit due to an indexing mistake, which has been fixed. Also, the Chebyshev model was actually using an order 1 less than requested because of python's non-inclusive maximum.
2. In `Spect.resampleToCommonFrame()` the combination `output_wave_scale="reference"` and `trim_spectral=False` is not allowed because it may require extrapolating the wavelength solution and round-tripping could become inaccurate. This combination of parameters previously raised a `ValueError` but not a warning is logged and `trim_spectral=True` is enforced.

### DIFF
--- a/geminidr/core/parameters_spect.py
+++ b/geminidr/core/parameters_spect.py
@@ -8,8 +8,12 @@ from astropy.io import registry
 from astrodata import AstroData
 from geminidr.core import parameters_generic
 from gempy.library import config, astrotools as at
+from gempy.utils import logutils
 
 from . import parameters_preprocess
+
+
+log = logutils.get_logger(__name__)
 
 
 def list_of_ints_check(value):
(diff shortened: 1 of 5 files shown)


KathleenLabrie:
Only one test is failing on Jenkins.

`test_resample_only` in `geminidr/gmos/tests/spect/test_resample.py` fails but I understand why.   The test is expecting an error to be raised, but we agreed to change the behavior to fix the setting instead of crashing.   Nothing serious.

Looks like that with the new behavior, the result should be the same as for `test_resample_only_and_trim`.    I'll try that.

chris-simpson:
Yes, sorry. I had spotted one test like this that I removed but obviously missed one. They were designed to check that an error was raised with an inappropriate combination of parameters, which you then convinced me should behave by changing one of the parameters.

KathleenLabrie:
No big deal.  Hopefully everything is green when you get back and you can take it from there.  (the install on jenkins failed in the "usual way", I just launch an new build.)

chris-simpson:
If the PR goes green and I'm not back at work you can just go ahead and merge it. There aren't any files that need moving around on Jenkins. Even if some of the other stuff that people have been working on has already been merged I don't expect any conflicts because this is pretty small and self-contained.

KathleenLabrie:
With `trim_spectral=True` forced when `output_wave_scale="reference"`, the test `test_resample_only` should behave exactly the same way as `test_resample_only_and_trim`.

I does not.  The w1 and w2 returned are the *untrimmed* values.   It appears that the `trim_spectral=True` is being ignored when it is reset within the code.

KathleenLabrie:
The PR is green on Jenkins.  Merging.

PR #487, opened 2025-03-27
A system/user-wide cache for calibration file downloads

phirstgemini:
This introduces an _optional_ system level cache for calibration file downloads. Most users will not need this and will not enable it. However it will make DRAGONS processing in GOA significantly more efficient.

If `system_calcache_dir` is defined in the `[calibs]` section of your `.dragonsrc`, then all calibration downloads from a `remoteDB` web server will be cached in this directory. If `system_calcache_dir` is not defined, no additional caching is introduced. For simplicity, no cache state is stored in the python code; cache hits and misses are purely on the presence and correct md5sum of files in the directory. I implemented this as a class to more easily allow implementation of more advanced cache features in the future should that become desirable, though in it's current form it could just be a function rather than a class if desired.

This is intended for use cases that will run several independent instances of DRAGONS in _different working directories_; either sequentially or simultaneously. Without this cache, each instance will download its own copy of the processed BPMs for example. GOA will do this extensively, which was the driver to implement this.

This PR also cleans up the `file_getter` code - eliminating for example file copies being implemented in bespoke python loops. 

This is completely independent of the "calibrations" subdirectory in the dragons working directory, which is referred to as a cache in the code. The cache implemented in this PR is in the code that downloads files to put them in those "calibrations" subdirectories.

### DIFF
--- a/recipe_system/cal_service/file_getter.py
+++ b/recipe_system/cal_service/file_getter.py
@@ -1,45 +1,112 @@
+import shutil
+import os.path
 import requests
 from requests.exceptions import HTTPError
 from requests.exceptions import Timeout
 from requests.exceptions import ConnectionError
 
-from gempy.utils import logutils
+from . import get_calconf
+from gempy.utils.logutils import get_logger
+from .calrequestlib import generate_md5_digest
+
+# get_file_itterator is the only function in this module called externally
+# (from calrequestlib.py)
+
+def get_request(url, filename, calmd5):
+    """
+    This function was moved from calrequestlib.py as it belongs more naturally
+    in this module as it is the only thing that calls methods in this module.
+    Those methods where replaced with the CachedFileGetter class.
+    :param url: URL to fetch data from
+    :param filename: Filename to put data into
+    :param calmd5: md5 hexdigest of cal we are requesting
+    :return: None
+    """
+    cfg = CachedFileGetter()
+    cfg.get_request(url, filename, calmd5)
+
+class CachedFileGetter(object):
+    """
+    This class implements a file getter with an optional system- (or user-)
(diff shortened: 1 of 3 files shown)


recipe_system/cal_service/file_getter.py
```diff
@@ -1,45 +1,112 @@
...
+                for chunk in r.iter_content(chunk_size=None):
+                    fp.write(chunk)
+            except HTTPError as err:
+                raise GetterError(
+                    ["Could not retrieve {}".format(url), str(err)])
+            except ConnectionError as err:
+                raise GetterError(
+                    ["Unable to connect to url {}".format(url), str(err)])
+            except Timeout as terr:
+                raise GetterError(["Request timed out", str(terr)])
```
teald, line 69:
```suggestion
            except HTTPError as err:
                raise GetterError(
                    ["Could not retrieve {}".format(url), str(err)]) from err
            except ConnectionError as err:
                raise GetterError(
                    ["Unable to connect to url {}".format(url), str(err)]) from err
            except Timeout as terr:
                raise GetterError(["Request timed out", str(terr)]) from terr
```

teald: Mainly suggesting this because it would capture the stack trace for the original error (in case it's relevant). I think leaving the message itself in the raised `GetterError` is fine, since it'll help inform in the original exception would be useful to look at.

teald (approved):
Looks good! Only had one minor comment, I'll approve now.

phirstgemini:
Thanks Teal, I've committed that suggestion. That code was copy-and-paste from the previous version, good catch to pass the underlying exception through, thanks.

PR #488, opened 2025-03-27
GNIRS longslit tutorials

KathleenLabrie:
This adds tutorials for GNIRS longslit.   There are 1 or 2 "todos", and maybe additional clarifications to be made but they work, the data gets reduced both in command line and API mode.

There are no changes to the code.  It's only documentation.


chris-simpson:
I'm approving this -- there was a small code change to a couple of regexes and then reversion.

PR #489, opened 2025-03-27
Fix GUI callbacks

chris-simpson:
There was an issue with incorrect callbacks when resetting the GUI reinit_panel. The telluric GUI uniquely has widgets with immediate callbacks to fit the data and when the "reset" button reset these, it resulted in multiple fits. I had fixed this in a way which I thought did not affect other GUIs but the textboxes weren't being reset in other GUIs. There were other commits in this area so there may have been unflagged conflicts or I may have messed up.

Anyway, I believe that the GUIs now operate as expected on hitting "Reset". The fitTelluric() GUI only performs one fit as a flag is set upon hitting "Reset" which causes the callbacks to no-op.

I also coerced the TELLABS spectrum from fitTelluric() to have the same dtype as the SCI plane. It's constructed from a spline call which defaults to 64-bit.

### DIFF
--- a/geminidr/core/primitives_telluric.py
+++ b/geminidr/core/primitives_telluric.py
@@ -241,7 +241,7 @@ class Telluric(Spect):
                 spline = make_interp_spline(tspek.waves[goodpix],
                                             absorption[goodpix], k=3)
                 spline.extrapolate = False  # will return np.nan outside range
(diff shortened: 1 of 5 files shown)


KathleenLabrie:
Tried it.  Confirmed fixed.

PR #514, opened 2026-04-18
Enh/check processed arc - add monitorWavelengthSolution primitive and associated qa recipes

phirstgemini:
Adds a `monitorWavelengthSolution` primitive (alongside determineWavelengthSolution) in `primitives_spect.py` that extracts some values from the WAVECAL table for the GOA instrument monitoring system to capture. I'm amenable to this living somewhere else but being in the same file and where the values are calculated and the .WAVECAL added seemed to make more sense. Also open to name suggestions if people don't like it.

Adds GMOS qa recipes that use that primitive that can be run by GOA to generate Arc fit statistics.

### DIFF
--- a/geminidr/core/parameters_spect.py
+++ b/geminidr/core/parameters_spect.py
@@ -579,6 +579,8 @@ class maskBeyondSlitConfig(config.Config):
         "Minimum fraction of pixel that must be illuminated to not be masked",
         float, 0.9, min=0., max=1., inclusiveMax=True)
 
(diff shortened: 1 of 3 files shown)


chris-simpson:
This seems reasonable, as per our discussions, although `_wavelengthSolutionMonitored` is an extension more in line with our existing ones. Or maybe "record" rather than "monitor" since monitoring is the thing you do with a sample (or with a single thing if you're going to act).

PR #515, opened 2026-04-20
Store SATLEVEL & NONLINEA consistently between primitives & NumPy 1 & 2

jehturner:
These keywords were being rounded differently by `ADUToElectrons` & `nonlinearityCorrect` and were differing by >0.1 on NumPy 1 & 2 (for SCORPIO) because the former calculates the mean less accurately with 32 bits. They were also getting propagated in memory as a NumPy dtype, which can cause comparisons to differ from the float produced when re-reading the file. This change should always store the values accurately to 3 DP as Python floats. See also 0c90e29.


jehturner:
Tests passed before I stopped a repeat run on the branch when merging.

PR #518, opened 2026-05-06
Fix regression when using alternative URL in download_from_archive()

jehturner:
Only attempt the new MD5 check from f270d67 when downloading from the Gemini archive, not alternative URLs, to avoid a `JSONDecodeError` (in the SCORPIO tests) when the expected URL isn't found.


jehturner:
Force-pushed the last commit with a new hash to work around a failure due to Jenkins forgetting the existing test results for the last commit on the branch (I think).

